PR #3052, opened 2015-01-28
Romanian almost full translation

titus08:
As soon as possible due to less available time that I have, I will translate the rest of 60 strings.


el-j:
updated files by hand. thanks

PR #3053, opened 2015-02-11
Full translation

titus08:
Full romanian translation.


aknoerig:
Awesome, Titus, thanks a lot!

titus08:
It was a hard work indeed. But since I am electronist and I use Kubuntu....

Issue #2445, opened 2014-08-26
display the selection width/height in the status bar if the mouse is dragged


davidperrenoud:
_From [irasc...@gmail.com](https://code.google.com/u/104729248032245122687/) on March 04, 2013 17:06:45_

http://fritzing.org/forum/thread/1412

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=2447_

PR #3056, opened 2015-02-22
Fixes fritzing/fritzing-app#2445

Snoozerworks:
Show width and hight of a mouse selection area in status bar.
More precise alternative to ruler part for simple part measurements.


aknoerig:
Very nice and good work, thanks!

PR #3074, opened 2015-04-14
polish translation - more elements

jacekjaros:
around 5% done


aknoerig:
Thanks, just keep it coming

PR #3083, opened 2015-05-13
Improvement to ground plane and gerber generation

nraynaud:
Hi, I added the clipper ( http://www.angusj.com/delphi/clipper.php ) library, and generated proper polygon boolean operation to avoid ragged edges to ground planes. I had to change the gerber generation too, because it was also creating ragged edges. I suggest you also use clipper for DRC checks.

I have increased the output resolution about 10 times, but if you zoom hard enough on the generated gerber, you'll still see the polygonized circles. I have created the ground plane by doing an erosion of the keepout size followed by a morphological opening ( http://en.wikipedia.org/wiki/Opening_%28morphology%29 ), then I cut the thermal isolation pads around the ground connectors.

I did my best not to break the panelizer, but I couldn't test it since, there is no documentation, nobody answered on my ticket, and subscribing to the forum is broken.

I hope you accept this PR, because it tremendously improves the quality of the output.

before 

after

It's interesting for people using CNCs to do isolation milling since the ragged edges are considered turns and the tool needs to stop before each turn (the toolpath is blue on the picture) and milling the PCB is needlessly slow.

The PNG logo import still uses ragged edges, because it was the simplest way to do it.

Nico.


aknoerig:
That's really great, thanks a lot! We always wanted to find a solution like that, since the scanline approach we took so far is obviously limited.
@el-j , can you test this on the panelizer process?

el-j:
after the last commit the dependent file svg2gerber.cpp was deleted. i'm unable to build with this commit.

nraynaud:
Yes, I deleted this file, I might have missed a reference, I'm re-checking for dangling references, it might take a couple hours because of the build time.

el-j:
ok, no problem. thank you for your great initiative!

nraynaud:
Ok, I think I nuked the file completely.

el-j:
That sounds great. I will test it at least till Sunday and let you know if
everything works. thank you very much

---\\> on the road

el-j:
hej again, i checkout the changes and sadly now there is problem with some QPointF *seedPoint
in the struct GPGParms. It's a C2864. Can you compile with your last commit?

image

nraynaud:
Shit, I guess my compiler is not checking hard enough. I guess this is allowed in c++11. I'm at a wedding so I can't check that right now.

nraynaud:
I hope it's fixed, I found another instance of the same notation. I don't have a windows machine to check myself.

el-j:
hej nraynaud. i have the same error now with some double in RectAperture and RoundAperture. 
if i delete the " = 0;" the error is gone and the only other error i get is a C2668 with some uint qHash. 

sadly i cannot find a way how i could fix it. 
thanks again for your support

el-j:
as addition: i use qt 5.2.1 and msvc 2012  32bit.

nraynaud:
Ok, I hope it works this time.

el-j:
ok, that solves the first problem. but the qHash problem seems to stay. what version of qt do you use for the build? i see some *if QT_VERSION code*. maybe it does not take this serious? 

anyway i downloaded qt5.4.1 to try the build with it. and IT WORKS. And it is amazing how fast
the copper-generator works now. and the edges are really  smooth.

but there are some bugs now and a lot of unsigned parameters.

**the gerber export:**
1. has a problem with smd-parts (the green circles)
2. it connects two donuts that was not connected in fritzing (red circles)
image
3. the drill file cannot be loaded into the gerber-viewer. error-message:
_hole parameters found inside programm-header. Programmheader (M48) must end with  "%"_
i try to add the % manually. i can import the drill file now but the drills are somewhere else ...

**fritzing bug:**
the mouse-cursor seems to be off.

have a look at the cursor if it is on "empty" space. (the normal white cursor)
and move it on an e.g. part/trace/pcb -> it seems: the cursor jumps to the left upper corner
of the white cursor. the clickable area is now at the right bottom corner of the cursor.

i'm unsure if the cursor-bug comes from qt5.4.1 or your changes.

nraynaud:
Can you send me the exported gerber and SVG please ? I'll have a look at the drill file during this time, I guess I missed a detail.

el-j:
here are new gerber and svg data together with the fzz. (i didn't save the first one :( )
the smd-part is ok now. but some things are connected as well. 

image


i already solved the cursor bug. it was a problem between qt5.2 and 5.4

https://drive.google.com/file/d/0By1pjKSYE4UkUU5rclBkNTVLaXc/view?usp=sharing

el-j:
after your last commit the drill file can be loaded but everything is extremely off.
(green pcb, red drill holes)

image

nraynaud:
ok, it's gonna get messy. In the gerber viewer I use, the file appear fine. I use GerbView, which is the gerber viewer from kicad. I have theory: the biggest change I made to the groundplane/gerber is that I added the support for polygons with holes (and since the groundplane generator created holes, I had to get a mean to export them, so I had to change the gerber exporter).

here is the file you sent me in Gerbview:
gerbview normal


To display holes in gerber I use dark and light polarity. 

here is a decomposition:
gerbview special
the white stuff is a clear polarity polygons that makes a hole in the ground plane, then there is the outline of all the traces and pads.

does your gerber tool allow for the display of light polarity?

el-j:
i can separate all copper top:
image

but i intend that this is not fine for all users. normally it should look like in your viewer. i use viewmate and there gerber seems to be broken (without any editing the gerber) and as far as i know, it is really complex to fix the gerber-output. do you have an idea how this thing happen?

nraynaud:
I'm looking into the excellon file format, the spec is a bit messy, and it looks like a 10 of 100 factor problem, I try to understand what was meant to be output (GerbView displays the holes fine).

nraynaud:
ok, let's try to talk the same language I used this specification for the format: http://www.ucamco.com/files/downloads/file/81/the_gerber_file_format_specification.pdf 

I created a gerber parser from this specification, and I can see the file fine in it too, so I might have a "personal" misunderstanding of the polarity system, and made the same mistake twice (in my parser, and in the Fritzing generator).
My understanding is that light polarity punches a hole into everything that is dark polarity. So I draw the outsidemost polygon in dark polarity, then I punches his holes in light, then I draw the islands in dark, etc.  recursively.

I'm trying to find some other mac compatible gerber software, so that I can try to slowly debug it.

aknoerig:
hi guys, i know that some people have said that our excellon drill file has always been off by a factor of 10 or 100 -- this doesn't show in all viewers, and usually no problem for production.
@el-j, maybe check with our manufacturing partner if they are alright with this.

el-j:
@aknoerig that doesn't make sense because if i cannot check the gerber output, i cannot give it to production.

aknoerig:
That's true, but they might at least be able to give us a hint and tell us what exactly the factor is by which it is off. :)

el-j:
from my point of view we have 2 main issues here: 
1. the drill file output
2. the svg2gerber output for the copper layers

the excellon export is issue 1. isn't it?

aknoerig:
Yes, excellon is the drill file format. I'm basically saying don't worry about this too much yet, as it might not really be an/the issue.

el-j:
@nraynaud i made some little export tests. what i found out:
without cooper fill the export is correct (right side).

image

so maybe there is just a "if copper-fill" problem in your gerber-generator?

nraynaud:
Sorry, my computer broke tonight, but i am committed to make this work as soon as i can.
To sumarize :
- fix the #if for qt version
- understand the problem with light polarity in gerber (found another application that incorrectly displays it)
- fix the factor in the excellon file

Envoyé de mon iPad

nraynaud:
I think I found an improvement to the generated gerber so that it is more compatible (I tested with 3 readers). I don't think what I did was wrong before, but I can see why other implementation might have missed a detail.

nraynaud:
Ok, I think I'm done with this wave of changes, could you try again please?

el-j:
AMAZING! 
image


i will do some more test, if they do not fail, i merge all the changes!

aknoerig:
Really cool! @nraynaud , now that you've come so far, would you like to try to apply this to the DRC, too? It's using the same algorithm as the copper fill. :)

nraynaud:
I would like to limit the scope of my intervention for now, I spent a few weeks on that already and I have other projects awaiting.

aknoerig:
Totally understood, thanks again for your investment! You're welcome back any time. We will probably issue a new release based on this feature. Contact me directly if we can mention your contribution in a special way.

el-j:
i'm really sorry, but my testing sadly bring up more failures with the export.

1. (pink circles): things outside of the pcb-board should be "deleted/clipped".
the svg file has a bigger size than its viewbox. i intent we have to specify that the gerber export strictly use the viewbox instead of the width and height of the svg. or the svg export should clip the things outside of the pcb-border. i'm not sure what i easier.
```
<svg xmlns="http://www.w3.org/2000/svg" y="0in" width="1.30603in" xmlns:svg="http://www.w3.org/2000/svg" version="1.2"  x="0in" viewBox="0 0 94.0343 103.219" height="1.43359in" baseProfile="tiny"/>
```
2. (green ellipses): really small "regions" are messed up or completely missing

image

image


here is my testpcb as fzz, gerber and svg export:
https://drive.google.com/file/d/0By1pjKSYE4UkN2pUTTNpYWhqT2M/view?usp=sharing

nraynaud:
Ok, I'll do the clipping right now.

I think the 2) was there before, when we load fzz files, "sometimes" they are messed up. This problem with the pads is also random and sometimes appears when loading the fzz file, so I deemed it as "outside the scope", to avoid having to debug the whole application. I'll give it a second look, but I'm not sure I'll fix it.

nraynaud:
I would like a bit of clarification for the clipping, please.
 - I clip the solder mask to the PCB outline
 - I clip the silk screen to the solder mask
 - I clip everybody to the PCB outline.
 - the solder paste stencil is the inverse of the solder mask

is that correct? am I forgetting something ?
Thanks, 

Nico

el-j:
hej again, i think you are right with the clipping. for me it sounds good what you write.

number 2)
we have this "jumping/moved wire bug" in fritzing. but we do not have a problem with the gerber export for smd parts and pads.

nraynaud:
sorry I'm a bit slow. I think the clipping works, the code is more and more arcane. I'll have a second look at 2) now.

el-j:
first of all: no, need to hurry! it's great that you do so much to develop fritzing!

your changes working like charm ;)   except for the drill file. 
could you clip the drill-file as well, please.

but bug number 2) seems to fixed as well. (till i can say now)
thank you

image

nraynaud:
strange, I'll give it a look, I gave a quick look to the drill file and thought someone else had done it :)

what should we do about the holes that partially intercept the PCB ?

nraynaud:
ok, I clipped every hole that would be at least partially outside the PCB. There is no clipping in the panelizer, because it calls back everything one layer at a time, so when I receive the order to generate the drill file, I don't have the information about the PCB shape (ditto for the silkscreen/soldermask clipping).

nraynaud:
any news about this PR?

el-j:
hej, i checked the changes and the normal fritzing-export works very good. the gerber seems to be correct and so smooth !!! ;) 
anyway, i cannot merge your request because the panelizer sadly crash the hole fritzing application now. i will try to find a solution for that. maybe we can make something like an "if else" so the panelizer still uses the old style...

nraynaud:
Well, I'm leaving for a trip tomorrow, but could you send me an example file to use the panelizer so I can fix it next month?

el-j:
yesterday i give it another try. these time the panelizer does the panels without a crash. but for some reasons the exported gerber has many broken pcb's. i will do some more testing to give you a better feedback. have a good trip. i'll answer with pictures just in time.

el-j:
sorry for the delay.

the panelizer feature is in a very experimental status. there are at least 2 different steps to use it.
and a bit of preparation before.

first of all you need a folder structure like:
`folder/orders`
`folder/products`

put the fzz's in a subfolder of the "orders" folder. this is your working directory. now you need an xml file with all the sketchs from the working-folder. the xml file has to sit in the same folder like the fzz's and have to look like:

```
<panelizer border="0mm" spacing="6mm" prefix="2014.09.09">
    <panels>
        <panel c2="65" c1="395" name="big" height="330mm" width="550mm"/>
        <panel c2="65" c1="297" name="small" height="310mm" width="225mm"/>
    </panels>
    <paths>
        <path>../../products</path>
    </paths>
    <boards>
   <board name="fritz_NewKeyfob.fzz" requiredCount="5" optionalPriority="0" produced="0" maxOptionalCount="0"/>
    </boards>
</panelizer>
```
*panel c1 is the amount of costs for the first panel, c2 for the copy of the panel*

*the board members are the fzz-files sitting in your working folder. they can have a required count or an optional count. the priority is for the optional boards and how important they are*

now you can use the inscriber. therefore you start fritzing with the -i option. (i for inscribe)
there are two more optional options: -nm (no message) and -drc (for an automatic drc check of the boards)

```
-i -nm "path to your panelizer.xml" -drc
```

after the inscriber checked all the files, it creates a copies folder, to save the original files. and it gives you an output text-file which fzz-files has failures. so now you can fix the failures. when you change anything in a sketch you have to inscribe again.

after that you can start the panelizer. therefor use the option -p (panelize) or -pc (panelize custom)
the pc option looks for sketches with custom created parts because they are often the reason for a bad gerber output. so, start fritzing with this option will begin the process:

```
-p "path to your panelizer.xml"
```

after the panelizer is done, you will have a gerber-folder in your working directory and a message from fritzing that the job is done. 

... i can see at the moment that the drill-data for most of the "mounting-holes" is missing.
and some things outside of the board-outline is exported as well. but i will do more testing during the next days.

thank you very much for all the good work till now!

Kazade:
Is there any update on this? It seems like this would be a huge improvement to the Gerber export. What's left to do to make this PR mergable?

nraynaud:
The panelizer doesn't work with this change.

I just lost interest, and I just use Eagle.

Kazade:
@nraynaud that's a shame - your work looks outstanding! If I can find time I'll pick up where you left off.

el-j:
i think this pullrequest is obsolete and all changes are merged with the pull-request #3203

thanks for this amazing work @nraynaud and @Kazade

PR #3102, opened 2015-07-09
Align to locked items (if any), otherwise align as usual.

canadaduane:
This allows users to align items to one or more locked items. For example, if the user locks the PCB in the background, and then selects several components INCLUDING the locked PCB, all components will be (left | right | center) aligned to the PCB. If no selected items are locked, the aligning functions work as before.


canadaduane:
Video explaining the feature: https://vimeo.com/133045425

aknoerig:
Very nice refinement, thanks! Especially since it doesn't introduce a new menu item, but just respects the lock state in an intuitive way.

PR #3111, opened 2015-07-12
Fix unicode 'degrees' symbol in Rotate menu

canadaduane:
At least on Mac OS, the 'degrees' symbol in the Rotate menu was showing up as a diamond-question-mark character. It appears the cause was a '\x' being used to describe unicode characters where '\u' would do.

https://dl.dropboxusercontent.com/u/118766/fritzing-unicode-before-after.png

### DIFF
--- a/src/items/partlabel.cpp
+++ b/src/items/partlabel.cpp
@@ -447,37 +447,37 @@ void PartLabel::initMenu()
 	bool include45 = (m_owner != NULL) && (m_owner->viewID() == ViewLayer::PCBView);
 
 	if (include45) {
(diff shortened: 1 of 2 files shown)


el-j:
hej canadaduane, i pulled your fix, but on win7 the degree symbols stays a question mark. so i did some research and find a stackoverflow: http://stackoverflow.com/questions/19550589/degree-unicode-showing-as-question-mark

so i changed the /u00B0 to the normal degree-symbol and everything is fine for me know. could you please check if that works for you too?
https://github.com/fritzing/fritzing-app/tree/canadaduane-fix_unicode

thanks

canadaduane:
See https://github.com/fritzing/fritzing-app/pull/3122

PR #3112, opened 2015-07-12
Add 'Wire' to distinguish 'Raise and Lower' menus

canadaduane:
I'm uncertain about the reason for having two "Raise and Lower" menus, but we should at least distinguish the visible names in the menu so they don't look like mistaken duplication.


canadaduane:
I just realized this also changes the right-click pop-up menu to read "Raise and Lower Wire", which may not be the desired effect? I'm a little confused as to why there are two "Raise and Lower" options in the main menu, and then the one (of those two possible options) that shows up in the right-click menu is the Raise and Lower Wire menu.

el-j:
i deleted the second "raise and lower" menu point. it seems to be a duplicate.
this change is already  included in development and master branch

PR #3122, opened 2015-07-26
Fix degrees symbols in Mac OS, retain in Windows

canadaduane:
Based on your suggestion, this should fix the degree symbol in all operating systems. (To be clear: I tested this in Mac OS and it looks good).


el-j:
thank you very much

PR #3123, opened 2015-07-26
Fix 'Fit in Window' regression

canadaduane:
This fixes the Ctrl-0 (Cmd-0) "Fit in Window" issues that the "Cursor-centered zooming, trackpad support on Mac OS X" patchset caused.

Note: I think this also fixes the other issues you were seeing (i.e. zooming factors out of sync among views, etc.)

### DIFF
--- a/src/sketch/sketchwidget.cpp
+++ b/src/sketch/sketchwidget.cpp
@@ -4337,8 +4337,8 @@ double SketchWidget::fitInWindow() {
 		itemsRect |= itemBase->sceneBoundingRect();
 	}
 
-    static const double borderFactor = 0.03;
-    itemsRect.adjust(-itemsRect.width() * borderFactor, -itemsRect.height() * borderFactor, itemsRect.width() * borderFactor, itemsRect.height() * borderFactor);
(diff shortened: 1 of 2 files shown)


canadaduane:
Any thoughts?

canadaduane:
If we had a faster dialog going, I could make improvements to Fritzing a little more quickly.

el-j:
sorry for the delay. i checked your changes. sadly the "zooming out of sync" stays. i merged your request to the development branch: 
https://github.com/fritzing/fritzing-app/tree/development

to get the issue:
1. zoom out to 5%
2. use mouse-wheel to zoom in again
3. the zoom-factor is much higher than befor. 
the result: on 264% zoom it is more like 2200% zoomed. zooming out to 5% again shifts up the zoomfactor again and again.

canadaduane:
It's strange, I'm trying to replicate this issue, but everything seems to work for me :( Perhaps it's because I'm on a Mac?

el-j:
with what qt version you are compiling? maybe we have some difference there?

PR #3149, opened 2015-10-16
Vietnamese Translation

duyetdinh:
Vietnamese translation


aknoerig:
This pull request is invalid -- there is no translation file here?

PR #3166, opened 2015-12-10
PCB ruler update

duff2013:
Whats new!
1. Ruler is now either Metric or English.
2. Metric has 1/100 Centimeter scale for fine measurement. 
3. Changing "Units" will be reflected immediately on window.
4. No global class members or variables added. Should not interfere with other UI i believe.

Below pics show the dynamic changing of units, also I changed the 1/100 centimeter scale to just one side:
screen shot 2015-12-10 at 4 49 12 pm
screen shot 2015-12-10 at 4 49 22 pm

One improvement I thought of was not showing the 1/00 centimeter scale until you are really zoomed in? 

I tried to keep the UI as simple looking as possible using the old Ruler as guideline. Let me know if there is anything you want changed, I hope my layout is acceptable to everyone?

### DIFF
--- a/src/items/ruler.cpp
+++ b/src/items/ruler.cpp
@@ -75,11 +75,18 @@ void Ruler::resizeMM(double magnitude, double unitsFlag, const LayerHash & viewL
 	double newW = TextUtils::convertToInches(QString::number(magnitude) + units);
 	if (w == newW) return;
 
+    // save local prop incase render fails so we can revert back to orignal.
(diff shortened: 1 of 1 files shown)


duff2013:
I have coded the change from using the combo box to radio button for the units:
screen shot 2015-12-13 at 12 19 04 pm
It is much easier to select the unit you want for the ruler with this change. I will submit pull once you decide to do with this pull request?

aknoerig:
Thanks, that's perfect! And yes, the radio button is preferred. :)

Issue #3132, opened 2015-08-18
Allow saving as non-zipped (fz instead of fzz) for version control?


SilverWingedSeraph:
Would it be possible to add the ability to save in a more git-friendly format? I.e., the fz XML format without the zipping.

ptaffet:
I also needed this feature, so I hacked it into the code. It's not particularly elegant, but it only took about 30 lines in FolderUtils::createZipAndSaveTo and FolderUtils::unzipTo that just mocked the zipping/unzipping and just copied the files directly without compressing.

aknoerig:
Nice, can you share the code? We could add it as a "save as...extracted fz" option.

ptaffet:
I forked it and committed these changes into my fork. My code works, but is quite ugly because I was trying to do it fast. The way the code in my branch works, there's a method at https://github.com/ptaffet/fritzing-app/blob/master/src/utils/folderutils.cpp#L345 that you can use to decide whether to create a real fzz or a fake fzz that just lists the files (then the files that would be in the fzz are saved directly in the same folder as the fake fzz). I was thinking that it could return true if the save destination is in a github repo or something like that, but to integrate this into actual production code, it should be a user decision.

zeehio:
This would be a nice to have feature!

* What should be done to convert the "ugly" code (your words, @ptaffet) :wink: into something mergeable for the next Fritzing release?
* Would @aknoerig review and accept a pull request if the patch is well written and meets quality requirements?

If @ptaffet is not available for tidying his patch I may try to to do it, although my experience with C++ and Qt code is a bit rusty/limited.

aknoerig:
@zeehio: absolutely, I'm happy to review!

ptaffet:
@zeehio My code is ugly, but worse, the way is exposed to the user is not acceptable for a production style. It should be an option in the Save As box. That requires more knowledge of Qt. I'm pretty busy this month, but feel free to use anything in my code. Two design points that need to be worked out are:
* How do you deal with the case when the zip would have multiple files? I'm not exactly sure when this happens, maybe if you include Arduino code. I solved this (IIRC) by storing in a folder.
* If the .fzz file becomes a folder, how do you open it? I solved this by keeping a "stub" fzz that caused Fritzing to look in the folder instead of unzipping the fzz to a temporary folder.

aknoerig:
Yeah, this can get tricky, so let's approach it pragmatically:

[...]

PR #3168, opened 2015-12-13
Load/Save as uncompressed fz file

zeehio:
This pull request starts dealing with issue #3132.

* Allow to save as *.fz file type:
  - Show "Fritzing uncompressed (*.fz)" in the "Save As" file type dialog.
  - saveBundledNonAtomicEntity function allows to save an *.fz 'bundle' by
    skipping the compression step (saves all files in the same directory).

* Allow to load a *.fz file:
  - Drop readonly warning.

* Rename private variable ___fritzingExtensions___ to p_fritzingExtensions as
  variables starting with underscore are compiler reserved in misc.cc.

TO DO:
  - Deal with overwriting existing auxiliary files. Maybe using a file +
    subdirectory approach would be safer?
     * mysketch.fz
     * mysketch.fz_FILES/
  - Should the save as uncompressed option be extended to other bundled types
    (such as uncompressed custom parts...)
  - Should loaded files be copied to another directory just like *.fzz files
    are extracted to another directory?
  - Thorough testing and code review. Saving and loading are delicate parts of the code and I do not want to be hold responsible of any bug...

### DIFF
--- a/src/mainwindow/mainwindow.cpp
+++ b/src/mainwindow/mainwindow.cpp
@@ -2475,12 +2475,7 @@ bool MainWindow::saveAs() {
     }
 
 	bool result = false;
-	if (m_fwFilename.endsWith(FritzingSketchExtension)) {
-		result = FritzingWindow::saveAs(m_fwFilename + 'z', false);
-	}
-	else {
-		result = FritzingWindow::saveAs();
-	}
+	result = FritzingWindow::saveAs();
 	if (result) {
 		QSettings settings;
 		settings.setValue("lastOpenSketch", m_fwFilename);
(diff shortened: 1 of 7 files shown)


el-j:
hello philip thanks for your work,
but i do not understand why you want to save an uncompressed fz. saving the fz will produce
a more complex behavior for all useres. e.g. parts can get lost very fast. what is the aim and the advantage for this option?

zeehio:
We (@ptaffet, myself and others) would like to save uncompressed fz files in order to be able to have a text-based format suitable for version control systems (such as git).

This change would allow easy Fritzing/Git (or github) interaction and would be very welcome by those of us who keep our arduino code in version control systems next to our schematics and documentation.

For sure we agree that for general releases, fzz files are the preferred way of saving a schematic (as you say parts are not lost).

I hope this makes sense to you, feel free to ask any question or discuss :-)

el-j:
ok, i checked your changes, and i really like how it works.
one thing is a bit "sad". when you look into the fz, all parts has absolute paths to the fritzing core paths.
in my case somewhere in my userfolder-structure of my machine. but i see that this is a problem of the fritzing-export. maybe we can find a way to change that, so fritzing uses the "-parts"-path as a variable. but i intent this is much more work. i will merge your branch after a bit more testing.

thanks for this idea and the work so far!

zeehio:
Thanks for considering it and merging it!

Issue #37, opened 2014-08-25
Icons for fritzing document types (.fz, .fzz, .fzb, .fzp)


davidperrenoud:
_From [dirk.van...@gmail.com](https://code.google.com/u/102845219688896911239/) on December 05, 2007 06:58:27_

We need document icons for the different document types that the
application can open and can create:

.fz    fritzing project file
.fzb   fritzing breadboard file
.fzs   fritzing schematic circuit file
.fzp   fritzing pcb (printed circuit board) view file

Those icons should be like the application icon is now, designed for
multiple sizes.

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=37_

aknoerig:
Only Linux still to do

el-j:
files are ready as svg and png. we still need to figure out how to set icons for file-extensions in linux (ubuntu). the icons has to be set via the mime. see:
http://stackoverflow.com/questions/30931/register-file-extensions-mime-types-in-linux

PR #3213, opened 2016-05-20
Added mac file icons

scribblemaniac:
Files associated with Fritzing (.fz, .fzz, etc.) now all have their own file icons on mac. The result looks something like this in the file browser:
screen shot 2016-05-19 at 8 34 08 pm

If you are testing this, you will likely need to use the deploy script. Additionally, the file icons may not update immediately because the icons are cached. If they don't update, just try copying some files around.

Related to #37, not sure if it closes it since I don't know what the status is for linux or window file icons.


aknoerig:
Awesome, thanks a lot!

scribblemaniac:
No problem!

PR #3224, opened 2016-06-07
Issues with fritzing.desktop

cjmayo:
One problem: that the Fritzing application is displayed without an icon on GNOME due to the rename of Icon value.

Plus some errors shown by desktop-file-validate.

More details on each commit. Happy to squash them into one.

```
$ desktop-file-validate fritzing.desktop 
fritzing.desktop: error: value "0.9.3b" for key "Version" in group "Desktop Entry" is not a known version
fritzing.desktop: error: value "icons/fritzing_icon.png" for key "Icon" in group "Desktop Entry" looks like a relative path, instead of being an absolute path to an icon or an icon name
fritzing.desktop: error: file contains multiple keys named "Categories" in group "Desktop Entry"
fritzing.desktop: error: value "Development;IDE;Electronics;EDA;" for key "Categories" in group "Desktop Entry" contains an unregistered value "EDA"; values extending the format should start with "X-"
fritzing.desktop: error: value "PCB;" for key "Categories" in group "Desktop Entry" contains an unregistered value "PCB"; values extending the format should start with "X-"
fritzing.desktop: hint: value "PCB;" for key "Categories" in group "Desktop Entry" does not contain a registered main category; application might only show up in a "catch-all" section of the application menu
```


aknoerig:
Thanks! How about if version=0.9.3, would that validate?

cjmayo:
Well coincidentally it would, but because 0.9.3 is an old version of the Desktop Entry Specification:
https://cgit.freedesktop.org/xdg/desktop-file-utils/tree/src/validate.c#n917

I think 1.0 is the only current possibility:
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-1.1.html#idm139653239401488
and it's optional anyway.

aknoerig:
Alright, thanks for the research!

PR #3231, opened 2016-06-12
use MIME database XML file and install_fritzing.sh

cjmayo:
I started this because I wanted to propose that a single fritzing.xml MIME database file is better to manage and package. Then realised it would break install_fritzing.sh, and ended up suggesting a number of changes to that.
Comments on individual commits.

I was tempted to include removing the `~/.mime.types` code. But maybe it was installed with a particular desktop environment or window manager in mind?

### DIFF
--- a/install_fritzing.sh
+++ b/install_fritzing.sh
@@ -1,7 +1,10 @@
 #!/bin/sh
 #
 # this is a rough beginning of a linux install script for fritzing
-# sets up document icons and file associations using mime types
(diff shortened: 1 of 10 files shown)


aknoerig:
That's great, thanks! I You're obviously much more versed in this than any of us (we only occasionally run Ubuntu in a virtual box). So you're welcome to take this further. If the mime.types code isn't required as far as you're aware of, let's remove it.

PR #3232, opened 2016-06-12
minimise initial size of cloned parts repository in releases

cjmayo:
Don't include the history or alternate branches.

### DIFF
--- a/tools/deploy_fritzing_mac.sh
+++ b/tools/deploy_fritzing_mac.sh
@@ -42,7 +42,7 @@ rm ./translations/*.ts  			# remove translation xml files, since we only need th
 find ./translations -name "*.qm" -size -128c -delete   # delete empty translation binaries
 
 echo ">> clone parts repository"
-git clone https://github.com/fritzing/fritzing-parts.git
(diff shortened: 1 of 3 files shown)


aknoerig:
Good point! I guess we didn't think of it because libgit2 doesn't support the --depth parameter. 
Did you test this with an update inside of Fritzing to make sure libgit2 doesn't get confused?

cjmayo:
No I haven't tried libgit2 from the GUI yet - I have problems with filepaths at runtime after install from source. I do have some ideas on that which are nearly ready to share!

cjmayo:
Good catch. It doesn't update from the GUI. libgit2 doesn't support shallow clones:
https://github.com/libgit2/libgit2/issues/3058

```--single-branch --branch master```  creates a repository that at the moment is only slightly bigger than the shallow one, but much smaller than a complete one (178M versus 293M).

I'll wait and see if this works from the GUI.

cjmayo:
Confirmed that works from the GUI. Commit updated.

aknoerig:
Great, thanks!!

PR #3233, opened 2016-06-12
make compatible with libgit2 >= 0.24.0

cjmayo:
Based on:
https://github.com/fritzing/fritzing-app/pull/3203/commits/1ffea750c05fda78c88b60b84545f6a3e1371159


aknoerig:
Thanks! There was also a request to make use of a system libgit2 if it exists, similar to how we treat boost, if you want to tackle that. This would require an update to phoenix.pro / utils.pri.

Issue #3237, opened 2016-06-16
sketches submodule


paulvollmer:
hey folk,
what do you think about replacing the [sketches/core](https://github.com/fritzing/fritzing-app/tree/master/sketches/core) directory by a submodule.

after i've create a new repository i realize that it make no sense if i create a submodule linked to my repo (wrong url at the submodule) and send a pull request...

if you're lazy, create a new [fritzing-sketches](https://github.com/fritzing/fritzing-sketches) repo i can fork and setup the submodule to the right directory  and move all the files to the new sketches repository.

failiz:
Is this still in the pipeline?

KjellMorgenstern:
Not planned.

PR #3238, opened 2016-06-17
Application Data and Parts Paths on Linux

cjmayo:
Something for discussion. Trying to solve issues having built and installed Fritzing 0.9.3b on Linux.

By installed I mean using make install (PREFIX=/usr).

Two issues:

1) If no parts library installed:

```
$ cd /
$ Fritzing
```

effectively hangs on the splash screen, because it is scanning the entire filesystem.

2) Where to put a GitHub clone of fritzing-parts that is user updateable? /usr/share/fritzing is read-only.

Suggestions:

Choose ~/.local/share/fritzing for the user clone, a directory already in the code.

Use PKGDATA to indicate building an installed Fritzing (would mean tools/linux_release_script/release.sh needs to do something different?)

At runtime look for fritzing-parts, then parts in each of (in order):
~/.local/share/fritzing
/usr/local/share/fritzing
$PKGDATA/fritzing

If none is found default to $PKGDATA/fritzing/parts. This still appears to be necessary to stop the scanning?

In this way a user can override a parts directory for themselves (with ~/.local),
also an administrator can override an installation for all users (with /usr/local).

N.B. application data directory and parts directory are independent in the local cases.

Do the same with the application directory for consistency. Will users want to install other types of files? Maybe:
https://github.com/fritzing/fritzing-app/issues/3237

Add a helper script, tools/user_parts_clone.sh, for users to create their own clone.

There may well be other ways to solve this. Maybe the directory scanning could be just for debug builds? But this seemed the simplest way to keep the current behaviour intact (OK looking in the local directories for the application path before $PKGDATA is a small change).

### DIFF
--- a/src/utils/folderutils.cpp
+++ b/src/utils/folderutils.cpp
@@ -118,6 +118,23 @@ QDir FolderUtils::getAppPartsSubFolder(QString search) {
 
 QDir FolderUtils::getAppPartsSubFolder2(QString search) {
 	if (m_partsPath.isEmpty()) {
+#ifdef PKGDATADIR
+		QStringList candidates;
+		candidates.append(QDir::homePath() + "/.local/share/fritzing");
+		candidates.append("/usr/local/share/fritzing");
+		candidates.append(m_appPath);
+		foreach (const QString &candidate, candidates) {
+			QList<QDir> dirList;
+			dirList.append(QDir(candidate + "/fritzing-parts"));
+			dirList.append(QDir(candidate + "/parts"));
+			foreach (const QDir &dir, dirList) {
+				m_partsPath = dir.absolutePath();
+				if (dir.exists()) {
+					goto setpath;
+				}
+			}
+		}
+#else
 		QDir dir = getApplicationSubFolder("fritzing-parts");
 		if (dir.exists()) {
 			m_partsPath = dir.absolutePath();
@@ -128,9 +145,10 @@ QDir FolderUtils::getAppPartsSubFolder2(QString search) {
 				m_partsPath = dir.absolutePath();
 			}
 		}
+#endif
 	}
 
-
+setpath:
 	QString path = search.isEmpty() ? m_partsPath : m_partsPath + "/" + search;
 	//DebugDialog::debug(QString("path %1").arg(path) );
 	QDir dir(path);
--- a/src/utils/folderutils.cpp
+++ b/src/utils/folderutils.cpp
@@ -74,6 +74,7 @@ QDir  FolderUtils::getApplicationSubFolder(QString search) {
 	path += "/" + search;
 	//DebugDialog::debug(QString("path %1").arg(path) );
 	QDir dir(path);
+#ifndef PKGDATADIR
 	while (!dir.exists()) {
 		// if we're running from the debug or release folder, try go up one to find things
 		dir.cdUp();
@@ -85,7 +86,7 @@ QDir  FolderUtils::getApplicationSubFolder(QString search) {
 
 		dir.setPath(dir.absolutePath() + "/" + search);
 	}
-
+#endif
 	return dir;
 }
 
@@ -215,8 +216,15 @@ const QString FolderUtils::applicationDirPath() {
 #ifdef Q_OS_WIN
 		m_appPath = QCoreApplication::applicationDirPath();
 #else
-		// look in standard Fritzing location (applicationDirPath and parent folders) then in standard linux locations
 		QStringList candidates;
+		// Look in standard Linux user and local administrator locations
+		candidates.append(QDir::homePath() + "/.local/share/fritzing");
+		candidates.append("/usr/local/share/fritzing");
+#ifdef PKGDATADIR
+		// look in installed location
+		candidates.append(QLatin1String(PKGDATADIR));
+#else
+		// look in standard Fritzing location (applicationDirPath and parent folders)
 		candidates.append(QCoreApplication::applicationDirPath());
 		QDir dir(QCoreApplication::applicationDirPath());
 		if (dir.cdUp()) {
@@ -229,13 +237,8 @@ const QString FolderUtils::applicationDirPath() {
 			}
 		}
 
-#ifdef PKGDATADIR
-		candidates.append(QLatin1String(PKGDATADIR));
-#else
 		candidates.append("/usr/share/fritzing");
-		candidates.append("/usr/local/share/fritzing");
 #endif
-		candidates.append(QDir::homePath() + "/.local/share/fritzing");
 		foreach (QString candidate, candidates) {
 			//DebugDialog::debug(QString("candidate:%1").arg(candidate));
 			QDir dir(candidate);
--- a/tools/user_parts_clone.sh
+++ b/tools/user_parts_clone.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+#
+# enable user to override installed parts library with GitHub clone
+
+REPODIR="${HOME}/.local/share/fritzing/fritzing-parts"
+mkdir -p "${REPODIR}"
+git clone --branch master --single-branch https://github.com/fritzing/fritzing-parts.git "${REPODIR}" || exit 1
+Fritzing -db "${REPODIR}/parts.db"
--- a/tools/deploy_fritzing_mac.sh
+++ b/tools/deploy_fritzing_mac.sh
@@ -19,6 +19,9 @@ builddir=$workingdir/../release64  # this is pre-defined by Qt
 echo ">> build directory"
 echo "$builddir"
 
+# not building for installation into filesystem
+sed -i 's:PKGDATADIR=\\\\\\"\$\$PKGDATADIR\\\\\\"::' phoenix.pro || exit 1
+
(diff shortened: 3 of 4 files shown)


aknoerig:
Thanks, good idea! I'm not sure I understand it completely, though. What would the install steps be now? And where is PKGDATA set?

cjmayo:
No changes to the existing installation process. A distribution can still package as before and also install parts in /usr/share/fritzing.
A user can choose to run user_parts_clone.sh to have their own GitHub parts library.

In the first comment I meant PKGDATADIR not PKGDATA. The variable in phoenix.pro.

To keep the same behaviour of the binary release (e.g. fritzing-0.9.3b.linux.AMD64.tar.bz2) PKGDATADIR should be undefined for building it.
The easiest would be to add ```sed -i "/PKGDATADIR =/d" phoenix.pro``` to release.sh and deploy_fritzing_mac.sh. Although that causes a number of warnings to be output when qmake is run - but they are harmless if you are not going to run ```make install```.

I'm starting to think that FolderUtils::getAppPartsSubFolder2 should have:
```
candidates.append(m_appPath)
```
instead of
```
candidates.append(QLatin1String(PKGDATADIR))
```
else it would break the use of ```-f <FOLDER>``` i.e. if you use ```-f``` you probably have to use ```-p``` as well.

cjmayo:
Updated:
- Use m_appPath instead of PKGDATADIR
- Change user_parts_clone.sh to use --single-branch

cjmayo:
@michaellass could you take a look at this?

I see in Arch you have worked around with:
```
--- a/src/fapplication.cpp
+++ b/src/fapplication.cpp
@@ -368,6 +368,7 @@ bool FApplication::init() {
 	m_serviceType = NoService;
 
 	QList<int> toRemove;
+	FolderUtils::setAppPartsPath("/usr/share/fritzing/parts/");
 	for (int i = 0; i < m_arguments.length(); i++) {
 		if ((m_arguments[i].compare("-h", Qt::CaseInsensitive) == 0) || 
 			(m_arguments[i].compare("-help", Qt::CaseInsensitive) == 0) || 
```
https://aur.archlinux.org/cgit/aur.git/tree/0003-Provide-a-sane-default-for-parts-path.patch?h=fritzing

michaellass:
I introduced this patch because without it the parts were not found at all. Looking at the current version of folderutils.cpp it seems like they are expected in a subfolder "fritzing-parts". At the same time "make install" installs the library into a subfolder called "parts" if it is placed under that name in the source directory before installation. So there is at least an inconsistency.

I think the logic proposed by this PR makes sense in order to allow installing a system-wide library as part of a package but allowing users to make use of the auto-update functionality by installing an own copy in their home directory. I tested the patches on top of current master as well as the script and they work fine.

One minor thing: The script should probably be marked executable in git.

cjmayo:
Thanks for testing.

It's true all the other shell scripts are executable in git, that's easily fixed (pushed with a rebase on current master). I wrote the script thinking of it as a sort of documentation - I guess it would be ideal to be able to do this from the GUI.

cjmayo:
Added sed's to tools/deploy_fritzing_mac.sh and  tools/linux_release_script/release.sh.
To remove adding PKGDATADIR to DEFINES rather than initialisation as suggested above.

michaellass:
Since my simple patch showed to introduce issues, I now provide a build including this PR to Arch Users. Feedback is positive so far. Is any help required to get this merged?

miqlas:
Please, use the XDG paths, not everybody got /usr!

heinervdm:
Perhaps I should comment on this, as the Linux default paths and PKGDATADIR was introduced by me.

In the beginning Fritzing was always looking for the data directories in sub folders of the current working directory, which did not work well when packaging it for a Linux distribution. As I didn't want to change to much of Jonathans code, I've just added the two locations I assume are used most often and in case someone wants to install it to a different location I've introduced the PKGDATADIR variable which allows the user to install the data files to any location he wants.

The main problem with this issue is, that qmake has no standard method to install applications, as cmake or autotools, at least I couldn't find one back then and therefore used method currently in use...

Because the issue @michaellass observed, I've just sent a pull request which will fix this issue.

cjmayo:
There are several issues here - I haven't checked but what is left to do after #3311?
e.g. the searching through the directories - I know I don't want that so putting it in `ifndef PKGDATADIR` was easy but can that be ripped out?

On not hardcoding `/usr` (which is already in the code before these patches) - moving to Qt5 only would allow the use of [QStandardPaths](http://doc.qt.io/qt-5/qstandardpaths.html) which might be a solution (not everyone is using an OS that supports XDG!).

heinervdm:
What one should keep in mind is: how can we make it possible that users that install Fritzing with their package manager profit from the self updating parts repository? When parts are installed in /usr or any other location not writeable by the user this can not work.
So the first step must be that the priority of ~/.local/Fritzing/parts is higher than that of /usr/share.
Also it would be nice if the partsfilder can be in ~/.local and the other directories still in /usr/share, which is currently not possible.

KjellMorgenstern:
Please see issue #3454 about this. The current proposal is to use one single folder, that resides in a user directory.

src/utils/folderutils.cpp
```diff
@@ -117,6 +118,23 @@ QDir FolderUtils::getAppPartsSubFolder(QString search) {
...
+			dirList.append(QDir(candidate + "/fritzing-parts"));
+			dirList.append(QDir(candidate + "/parts"));
+			foreach (const QDir &dir, dirList) {
+				m_partsPath = dir.absolutePath();
+				if (dir.exists()) {
+					goto setpath;
+				}
+			}
+		}
+#else
```
KjellMorgenstern: I think those guesses are not required, and are quite OS specific.
The --parts CLI option can be used instead.

src/utils/folderutils.cpp
```diff
@@ -215,8 +234,15 @@ const QString FolderUtils::applicationDirPath() {
...
-		// look in standard Fritzing location (applicationDirPath and parent folders) then in standard linux locations
 		QStringList candidates;
+		// Look in standard Linux user and local administrator locations
+		candidates.append(QDir::homePath() + "/.local/share/fritzing");
+		candidates.append("/usr/local/share/fritzing");
+#ifdef PKGDATADIR
+		// look in installed location
+		candidates.append(QLatin1String(PKGDATADIR));
+#else
+		// look in standard Fritzing location (applicationDirPath and parent folders)
```
KjellMorgenstern: Same as above.

tools/user_parts_clone.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+#
+# enable user to override installed parts library with GitHub clone
+
+REPODIR="${HOME}/.local/share/fritzing/fritzing-parts"
+mkdir -p "${REPODIR}"
+git clone --branch master --single-branch https://github.com/fritzing/fritzing-parts.git "${REPODIR}" || exit 1
+Fritzing -db "${REPODIR}/parts.db"
```
KjellMorgenstern:
```suggestion
Fritzing -parts "${REPODIR}"
```

src/utils/folderutils.cpp
cjmayo: Yes these patches (and others at the time) were about getting 0.9.3b working on Linux. However, the paths are just those that are already in the code. This was just a way of rationalising their use - especially being able to avoid the directory recursing with 3724a283c88a5ed964ec6f23beddbb273ddbdda0.

I guess the other option is to remove all the code that is trying to find the parts in different locations, maybe leaving only applicationDirPath and PKGDATADIR (haven't thought about that too much).

tools/user_parts_clone.sh
cjmayo: This line is about generating parts.db, not starting Fritzing. I think that is OK?
I can split this commit off into a separate pull request if that would be useful.

KjellMorgenstern: A script to switch part repos quick would certainly be useful for parts development, so yes, you could split that out (maybe even move to the fritzing-parts repository) If you just change the part files, not the location (like 'git checkout') then updating the -db would be needed. But I assume, when you run 'git clone', that you now have a directory that was previously not there, and to tell fritzing about this new directory, you would need the -path option, not -db.

cjmayo: OK. Makes sense. I think it needs a conclusion on what is going on with the directories before moving forward with this script - `~/.local/share/fritzing` would be supported at present.

src/utils/folderutils.cpp
```diff
@@ -117,6 +118,23 @@ QDir FolderUtils::getAppPartsSubFolder(QString search) {
 
 QDir FolderUtils::getAppPartsSubFolder2(QString search) {
 	if (m_partsPath.isEmpty()) {
+#ifdef PKGDATADIR
+		QStringList candidates;
+		candidates.append(QDir::homePath() + "/.local/share/fritzing");
```
gasinvein: Please don't hardcode `~/.local/share`. It's a fallback value for cases when `$XDG_DATA_HOME` env var is not set. See the [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

PR #3242, opened 2016-06-19
[WIP] Generate man page from improved --help and add --version

cjmayo:
Looking at the --help output I saw it was missing some options and also realised it would be straightforward to generate a man page using help2man that uses --help (and --version which is added here).

It does mean sending --help to stdout. I'm only testing on Linux. There are some code changes which should be reviewed here. Is the enum use, location and naming OK? In phoenix.pro `manpage.files` didn't seem to work for a generated file.

I've proposed truncating m_revision, don't know if that breaks anything else but it makes the man page neater!

Attached is a generated man page. Should be viewable with `man -l Fritzing.1.txt`
[Fritzing.1.txt](https://github.com/fritzing/fritzing-app/files/322483/Fritzing.1.txt)

Still a bit to-do here. I have put "???" to indicate things that definitely need doing, but needs carefully checking for accuracy and appropriateness too. I've taken some text from the website and readme.md.

I've made some of the --help output conditional on PKGDATADIR, which links this to #3238. Would be worth checking for any help changes after this anyway.

help2man is then another dependency for building from source for unix but is available on Debian/Ubuntu etc.
and looks like it can be installed on Mac even (I have no knowledge of this, so this is definitely not a recommendation):
http://macappstore.org/help2man/


cjmayo:
Rebased on develop.

KjellMorgenstern:
I am hesitant to include this, since adding those parameters to the help and man pages would mean to officially support them. Is there good reason for any of those parameters to be included? Is any of them already an essential part of common use cases?

cjmayo:
I think this was always one for comment and not immediate merging.

I guess either way each of the options need reviewing and if not wanted the actual code implementing them should be removed, not just the documentation.

My personal choices: 
*  `Fritzing  [FILE]` to be able to open a project directly from the command line.
* `-debug` does show a number of messages, and its easy to add to them while developing
* `-help` if you have any options this should be present
* `-version` great for help2man, see below

From memory I think while looking at the parts folder code I used the likes of `-parts FOLDER`, `-folder FOLDER`, `-database FILE`.

It will need more advanced users than me to review the others.

If a program has options it should have:
* `-help` written to stdout not stderr
*  a man page (requirement on Debian?).

help2man IMHO is a great way to only have to maintain documentation in one place, especially when there is already a `--help` and it's still actively maintained:
https://salsa.debian.org/bod/help2man

Perhaps we could split off some commits into other PRs:

1) remove "}]" from end of some lines in --help output
send --help to stdout

2) add --version
only use first 8 characters of Git SHA-1 checksum for revision

3) generate man page using help2man

Let me know if you would like any of these.

KjellMorgenstern:
Yes, please split. 
I think 1 + 2 can go in easily.

cjmayo:
We've worked through all the commits here that could be merged straightforwardly.

The two left would need more work:
- add undocumented options
- use help2man to generate man page (I expect also needs adding to docker image)

PR #3243, opened 2016-06-22
tools deploy_fritzing_mac.sh and release_fritzing.bat set clone depth…

paulvollmer:
… for speedup the script execution time


aknoerig:
I'm afraid this won't work. See https://github.com/fritzing/fritzing-app/pull/3232
Or does depth=2 instead of depth=1 make the difference?

PR #3248, opened 2016-07-03
Support System Boost and libgit2

cjmayo:
Here's a start on making phoenix.pro usable with system libgit2 and non-deb Boost.

**\* I've only tested on Linux with system Boost and libgit2.

First some tidy-ups on indentation etc. in phoenix.pro.

Moves the detection of out into separate .pri files to keep phoenix.pro as small as possible and make the testing for installed versions clear.

libgit2 detection is OK I think, thanks to its and qmake's support of pkg-config (I assume it doesn't run on macx?). As long as packagesExist() doesn't cause an error on win32 or macx.

The change for Boost just optionally skips the Boost detection and checking using a new BOOST_INSTALLED definition.

BOOST_INSTALLED constant isn't actually needed in the code (unlike QUAZIP_INSTALLED), so defining a SYSTEM_BOOST and a SYSTEM_QUAZIP to match might be better. On the other hand maybe BOOST_INSTALLED should be used in the code to force the use of the non-installed Boost version (although that may not be so simple).

A few ideas for boost detection that I haven't tried:
- adding compile time detection to graphutils.cpp, exiting with #error
- use qtCompileTest(test) in phoenix.pro

### DIFF
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -46,8 +46,7 @@ win32 {
         RELDIR = ../release64
         DEBDIR = ../debug64
         DEFINES += WIN64
-    }
-    !contains(QMAKE_TARGET.arch, x86_64) {
+    } else {
         RELDIR = ../release32
         DEBDIR = ../debug32
     }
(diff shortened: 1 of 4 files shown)


aknoerig:
Excellent, thanks! I'll give this a try on Mac and Windows.

aknoerig:
I had to adapt the libgit2 paths, since the .pri is now in a subfolder. Otherwise works fine!
I failed to push to my changes to this pull request though, and instead opened a new PR #3249 about to be merged.

Issue #3228, opened 2016-06-08
0.9.3b: Cannot use system boost anymore


michaellass:
I'm maintaining an unofficial fritzing package for Arch Linux: https://aur.archlinux.org/packages/fritzing/

When trying to package 0.9.3b, build fails with the following message:

> sh: dpkg: Command not found.
> Project MESSAGE: Please download the boost library--you can find it at http://www.boost.org/
> Project MESSAGE: Note that boost 1.54 has a bug in a function that Fritzing uses, so download some other version
> Project ERROR: Copy the boost library to .../src/lib/, so that you have .../src/lib/boost_1_xx_0

Boost is installed on the machine (version 1.60.0) and fritzing 0.9.2b is perfectly happy building against the already installed boost version. Is this an intended change or some regression? I found #2366 which leaves the impression that it should be possible to build against the system boost libraries.

Also I had to manually extract and build libgit2 as described in the wiki although libgit2 is installed on the system (in version 0.24.1 and not 0.23.4 as required according to the wiki, though…). Would it be possible to build against the system library here, too?

All in all I'd like to keep the redundancy of installed libraries as low as possible. Are there any switches I missed?

aknoerig:
We'd like to keep it that way, too! 

Regarding boost, it should still take the local version, and at least it does so on my ubuntu machine. We did make some changes to the include logic, maybe that's incompatible with archlinux?
See https://github.com/fritzing/fritzing-app/commit/2a97c2fb5e070dcf4a5bacdca021d58b9d53bdf2

Regarding libgit2, we'd need to add a few ifdefs since the API between 0.23.4 and 0.24.1 slightly changed. Could you give that a try yourself?
All git-related code is in https://github.com/fritzing/fritzing-app/blob/master/src/version/partschecker.cpp, and the include logic is here: https://github.com/fritzing/fritzing-app/blob/master/phoenix.pro#L164

michaellass:
Thanks a lot for these pointers! It's pretty clear that the boost include logic currently only works on distributions using dpkg. So also RedHat, Fedora, Suse, Gentoo etc. currently won't use the system boost. I will see if I can find another way to determine the version of boost installed, so no access to a package manager is required. As a first step I now know what to patch to allow building on Arch Linux.

I'll also have a look into the libgit2 API changes.

cjmayo:
[...]

PR #3250, opened 2016-07-04
Use ldconfig for boost detection

michaellass:
dpkg is only available on debian based systems. Use ldconfig to determine the installed version of boost.

This addresses #3228. I only tested it on Arch Linux with a recent version of boost installed. It should be tested on a debian based system to make sure it does not introduce a regression.

libboost_filesystem.so is an arbitrary choice -- it just needs to be one of the installed boost .so files.

### DIFF
--- a/pri/boostdetect.pri
+++ b/pri/boostdetect.pri
@@ -29,14 +29,12 @@ for(boost, BOOSTS) {
 contains(LATESTBOOST, 0) {
     unix {
         !macx {
-            BOOSTINFO = $$system(dpkg -s libboost-dev | grep 'Version')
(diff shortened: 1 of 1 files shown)


cjmayo:
ldconfig is in /sbin on Debian:
https://packages.debian.org/jessie/amd64/libc-bin/filelist
and Gentoo where /sbin is not in an ordinary user's path.

An alternative approach for discussion here https://github.com/fritzing/fritzing-app/pull/3251

michaellass:
Thanks for testing! Checking for boost using boost/version.hpp like in #3251 definitely is the preferred way.

PR #3251, opened 2016-07-04
check Boost version with qtCompileTest instead of dpkg

cjmayo:
Doesn't distinguish between no Boost and Boost 1.54. Could add a second test if this really matters.
Does this actually work on Mac and Windows?
Messages might be updated.


aknoerig:
Works fine on Mac, too, thanks!

PR #3252, opened 2016-07-06
add option to choose installed translations using LINGUAS

cjmayo:
Saves a lot of code in downstream packaging with selected translations only e.g.:
https://github.com/gentoo/gentoo/commit/53c843126785b377f6cbca9d2ea03001f9153dbd#diff-13ffa73c7e68e1ac2f92e7aa89c68cc2R50

### DIFF
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -143,7 +143,12 @@ unix {
     bins.files += bins
 
     translations.path = $$PKGDATADIR/translations
-    translations.extra = find $$PWD/translations -name "*.qm" -size +128c -exec cp -pr {} $(INSTALL_ROOT)$$PKGDATADIR/translations \\;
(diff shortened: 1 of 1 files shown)


cjmayo:
Maybe L10N isn't the best choice of name for the variable as the language names are in "language_country" form - with the country in lower case. And if they really are just translations and not conversion of units etc.

Any particular reason for the lower case country suffixes? (FApplication::findTranslator)

Selfishly BCP47 format with hyphens and upper case region subtag would be easier to handle. But I guess changing it could break a user's saved configuration.

cjmayo:
I think LINGUAS Is better, typically uses underscores instead of hyphens.

Upper case country codes would be great but easily converted with ```${LINGUAS,,}```

PR #3254, opened 2016-07-06
use SYSTEM_QUAZIP to choose installed QuaZIP

cjmayo:
Suggest using SYSTEM_QUAZIP=1 with qmake instead of DEFINES+= QUAZIP_INSTALLED, and testing that QuaZIP really is installed (halting if it isn't).

**\* haven't tested: tools/linux_release_script/release.sh

Also a tidy-up for system libgit2 and ensure messages from qmake are only shown once.

### DIFF
--- a/config.tests/quazip/main.cpp
+++ b/config.tests/quazip/main.cpp
@@ -1,0 +1,4 @@
+#include <quazip/quazipfile.h>
+int main()
+{
+}
--- a/config.tests/quazip/quazip.pro
+++ b/config.tests/quazip/quazip.pro
@@ -1,0 +1,2 @@
+CONFIG += qt
+SOURCES = main.cpp
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -102,7 +102,7 @@ unix {
         } else {
             DEFINES += LINUX_32
         }
-        !contains(DEFINES, QUAZIP_INSTALLED) {
+        isEmpty(SYSTEM_QUAZIP) {
             LIBS += -lz
         }
     }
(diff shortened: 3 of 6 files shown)


cjmayo:
QuaZIP 0.7.2 changed for Qt5 to installing:
/usr/include/quazip5/quazip.h
/usr/include/quazip5/quazipfile.h
/usr/lib64/libquazip5.so

Maybe this isn't such a good idea. A simpler one:
```
diff --git a/phoenix.pro b/phoenix.pro
index 26e25b5..58c5f2f 100644
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -199,8 +199,13 @@ include(pri/program.pri)
 include(pri/qtsysteminfo.pri)
 
 contains(DEFINES, QUAZIP_INSTALLED) {
-    INCLUDEPATH += /usr/include/quazip
-    LIBS += -lquazip
+    greaterThan(QT_MAJOR_VERSION, 4):exists(/usr/include/quazip5) {
+        INCLUDEPATH += /usr/include/quazip5
+        LIBS += -lquazip5
+    } else {
+        INCLUDEPATH += /usr/include/quazip
+        LIBS += -lquazip
+    }
 } else {
     include(pri/quazip.pri)
 }
diff --git a/src/utils/folderutils.cpp b/src/utils/folderutils.cpp
index 96f10c7..576e729 100644
--- a/src/utils/folderutils.cpp
+++ b/src/utils/folderutils.cpp
@@ -41,8 +41,8 @@ $Date: 2013-04-29 13:10:59 +0200 (Mo, 29. Apr 2013) $
 
 #include "../debugdialog.h"
 #ifdef QUAZIP_INSTALLED
-    #include <quazip/quazip.h>
-    #include <quazip/quazipfile.h>
+    #include <quazip.h>
+    #include <quazipfile.h>
 #else
     #include "../lib/quazip/quazip.h"
     #include "../lib/quazip/quazipfile.h"
```

KjellMorgenstern:
There are no quazip related build problems on the develop branch, so this is not urgent. If I am correct, the improvement here is that the build will immediately stop if quazip is not available, instead of failing later during linking. That would be nice to have, but the PR would have to be rebased, taking recent changes into account. Therefore, closing it for now.

PR #3276, opened 2016-09-17
New Updates in Czech language (17.9.2016)

Vasekdvor:
New Updates in Czech language (17.9.2016)
- Newly completely translated Tips & Tricks
- Newly completely translated First time Help (on each screen PCB,BreadBoard and schematics)
- Completely translated Preferences dialog window
- Some minor changes and corrections (Rotating, Autoroute menu, About)
- Background color dialog some translations (but it needs to be fixed and need some restructuralization)

BTW: Context menu where is Rotating objects is still not using translated strings
I mean "Context menu" -> "Rotate" and these five strings:
Rotate 45° Clockwise
Rotate 90° Clockwise
Rotate 180°
Rotate 90° Counter Clockwise
Rotate 45° Counter Clockwise
These strings is still not translated because the fritzing seems doesnt link these strings to czech (and probably any other language) file.
Because these strings is translated in czech file.


el-j:
this pull request had failures in the ts file. sadly the original string gets replaced by the translation. 
i fixed this manually. so this request is obsolete and closed. 

thx for the part translation. please take care how to use the ts file for translation.

PR #3290, opened 2016-11-26
Changing Linux Installer

GQDeltex:
Changing the Linux Installer.

The fritzing.desktop file is now copied to /usr/share/applications so it shows up in the app-launcher
Also I added a file in /bin/ so you can start fritzing by simply typing fritzing in the command line.
All the files are also moved to ~/.fritzing so it is more user-friendly (you don't notice it)

Thanks, GQDeltex

### DIFF
--- a/fritzing
+++ b/fritzing
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+cd ~/.fritzing-0.9.3b
+./Fritzing
--- a/fritzing.desktop
+++ b/fritzing.desktop
@@ -2,8 +2,8 @@
 Name=Fritzing
 GenericName=Fritzing
 Comment=Electronic Design Automation software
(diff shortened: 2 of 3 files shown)


KjellMorgenstern:
@GQDeltex Which linux distribution / standard did you have in mind or test this on?

cjmayo:
I think it would be interesting to have a bug report on the problems with the original script. At a glance a lot of these changes seem problematic. Indeed if you run the original script from within `~/.fritzing` it might achieve a lot of what is wanted (haven't tested).

The one thing the original script doesn't do is allow you run `fritzing` from the command line. Although a symlink in /usr/local/bin should do the trick. Indeed maybe a whole other install script targetting /usr/local would be an idea.

Issue #3012, opened 2014-10-02
Russian for desktop file


glixx:
Comment[ru]=Программа проектирования электроники

KjellMorgenstern:
This issue was closed because it is outdated.
If you think this issue is still important for a reason, please open.

glixx:
Not outdated.

PR #3295, opened 2017-01-01
Update desktop file

jnumm:
Caught by Debian's Lintian: https://lintian.debian.org/tags/desktop-mime-but-no-exec-code.html
"The desktop entry lists support for at least one mime type, but does not provide codes like %f, %F, %u or %U for the Exec key.
If the application can indeed handle files of the listed mime types, it should specify a way to pass the filenames as parameters."


jnumm:
This also includes "Russian for desktop file" #3012.

PR #3299, opened 2017-01-11
Metadata update: desktop entry, appdata

jnumm:
These commits update and extend the linux distro metadata files.
The specs are: https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html and
https://www.freedesktop.org/software/appstream/docs/index.html


ovidiub13:
@jnumm Please rebase this PR off of Fritzing-app `develop` branch. The changes look good so far, but I'd like to dive a little deeper into them. Also some of these files I'd prefer to have generated, so translations are all handled in a single place.

jnumm:
@ovidiub13 thanks for looking at this. The rebase to `develop` is now done.

I agree that having all the translations in one translation file would be much nicer but Qt Linguist tools don't seem to support that out of the box. For now, I don't think I'll be going to contribute a solution for generating this metadata.

My commits touch the file `install_fritzing.sh` in the relevant parts despite that script not being in a working state, for example it is trying to copy files from a non-existing path (https://github.com/fritzing/fritzing-app/blob/develop/install_fritzing.sh#L178).

KjellMorgenstern:
What is wrong about the translations? Usually, the translators and the developers are not the same person, and work at entirely different phases of the project, and in different langugages ;-)
If you mean the binaries...  they are checked in because we want to run translations once for all build targets (Windows, Linux, MacOs, various distributors who build their own binary). It is quite a manual process. For example, if you correct a typo in the source, you don't want to trigger a translation for all languages (the translator does not transfer the typo).

ovidiub13:
@KjellMorgenstern I was talking about the `.desktop` file. It has comments in several languages. I feel it might be better to have that file generated, so it doesn't have to be manually changed every time someone translates it to a new language.
I haven't thought this through, so IMO it's not a priority. We can look for a solution next time we touch this file.

PR #3311, opened 2017-02-04
folderutils.cpp: Fix getAppPartsSubFolder2 returns current working di…

heinervdm:
…rectory if fritzing-parts folder is not found

getApplicationSubFolder returns QDir() if the directory can not be found,
but QDir() returns the current working directory, which nearly always exists and probably is the wrong path.
As getAppPartsSubFolder2 only checks for existence of that directory we end up with the wrong directory if parts folder is used instead for fritzing-parts.
Therefore check if when we ask for fritzing-parts folder we really get a folder with that name.


cjmayo:
Is it possible instead to put these tests inside getApplicationSubFolder()? And for it not to return the home directory?

heinervdm:
I would find this nicer too and tried to implement it that way, but the result of getApplicationSubFolder is only checked in getAppPartsSubFolder2 and not in the other places it is used. So returning an non existing folder can cause problems in these cases. Also I found no clear way to indicate an error in the return value, when returning the QDir by value.

PR #3339, opened 2017-05-28
Allow headless builds on OS X

nopdotcom:
Deployment of Fritzing.app on OS X bombs out while trying to build the parts database. The crash happens inside `Fritzing -db foo`, which can’t set up its mouse pointers because it has no UI.

This patch skips does cursor initialization only if the service is `NoService`. Are there other conditions when we **do** want to bring up a GUI?

Tested very lightly.


nopdotcom:
This bug might not be exposed if Fritzing.app is being built from a Terminal directly on the build machine, as there would be a GUI available. I do my builds remotely, and talking to the display fails.

PR #3343, opened 2017-07-10
Fix SVG font handling for single-quoted font-family inside style attribute

historicbruno:
With latest InkScape and Fritzing releases, one receives an error loading a schematic SVG, as reported here: http://forum.fritzing.org/t/font-error-in-parts-editor-when-reading-schematic-svg-which-is-edited-with-inkscape/3753

The problem is the SVG font fixing handles single quotes in the font-family attribute, but not inside the style attribute. The minimal fix was to update the regex in `removeFontFamilySingleQuotes` accordingly to support both.

### DIFF
--- a/src/utils/textutils.cpp
+++ b/src/utils/textutils.cpp
@@ -118,7 +118,7 @@ FixedFontsHash fixFontsMapping(const QSet<QString> fontsTofix, const QString & d
 }
 
 bool removeFontFamilySingleQuotes(QString &fileContent) {
-	static QString pattern = "font-family=\"('[^']*')\"";
+	static QString pattern = "font-family(?:=\"|:)('[^']*')\"?";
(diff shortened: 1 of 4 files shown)


KjellMorgenstern:
Trying to reproduce the problem and did a quick grep for "family:Droid" , but the parts that I found and tried did not show the issue. Could you upload or refer a part? I will then try to reproduce the issue again.

historicbruno:
> 
> 
> Trying to reproduce the problem and did a quick grep for "family:Droid" , but the parts that I found and tried did not show the issue. Could you upload or refer a part? I will then try to reproduce the issue again.

Well I'm not sure if I was clear enough, but you would actually need to grep for "family:'Droid"  (note the single quote causing the problem, although I can't remember off the top of my head if there was also whitespace in there). I've not tested with the latest source code.

PR #3344, opened 2017-07-10
Fixes quazip build on Windows with recent versions of Qt

historicbruno:
I have Qt 5.8.0 installed on Windows 10, and it doesn't include a bundled QtZlib, causing the Fritzing build to fail on quazip. I think it was incorrect (though convenient) to rely on the past existence of QtZlib. Instead, I have modified the build scripts to "detect" zlib on Windows, following the model of `libgit2detect.pri`.  It assumes zlib is extracted and built in `src\lib\zlib`. I moved the *nix zlib configs to the new `zlib.pri` as well.

I have only tested on Windows, as I don't have a build env for Linux or macOS. NOTE: This requires an update to the build instructions, to direct the user to download zlib and build it using e.g. CMake -- very similar to the process for libgit2.

### DIFF
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -1,7 +1,7 @@
 # /*******************************************************************
 #
 # Part of the Fritzing project - http://fritzing.org
-# Copyright (c) 2007-16 Fritzing
+# Copyright (c) 2007-17 Fritzing
 #
 # Fritzing is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


vanepp:
For what its worth, your fix appears to work on a Ubuntu 16.04lts desktop install on Qt 5.10 as far as I can see (Fritzing builds and loads a part which should exercise the unzip library), so thank you, because I don't know enough about Qt to have done this. Hopefully that will also help me get it running on Visual Studio on Windows (which is being more exciting to build than linux) as I have some fixes to head I want to run there. I had to hand apply the patches because I couldn't figure out how to get github to give them to me to merge in to my branch, it keeps giving me the master branch as far as I can see but alls well that ends well . ..

Peter

KjellMorgenstern:
Rebased this as PR #3464 to see it in CI , not sure if this still applies.

PR #3354, opened 2017-09-24
Use delete[] where new[] was used for allocation

cmdrf:
These arrays were allocated with `new[]`, so they have to be deleted with `delete[]`. Otherwise you risk memory leaks.


mike168m:
Can they be replaced with a vector of smart pointers instead?

cmdrf:
Using a vector of smart pointers would have a significantly lower performance and increase code complexity. I would rather use `std::vector<E>` or `QVector<E>`, or use `std::unique_ptr` with a custom deleter. The latter requires C++11 however.

Issue #3333, opened 2017-04-25
NEW GerberOutput seems broken


el-j:
There is a problem with the new gerber output when the pcb has copper or ground-filling:

screenshot from gerbV:
![image](https://cloud.githubusercontent.com/assets/2795534/25391684/182e34d2-29d7-11e7-9ede-9b1b9160ff7d.png)

it seems that it is a problem with circles, they are not closed properly:

![image](https://cloud.githubusercontent.com/assets/2795534/25393630/40e7776c-29dc-11e7-9e74-b6556868ce86.png)

@nraynaud @Kazade do you have any idea why this happens?

if i look at the gerberout in a text editor i cannot see any circle commands G02/G03 or G75,
shouldn't there be just 2 circles in the last screenshot?

thanks for your help! 

```
%ASAXBY*%
%FSLAX26Y26*%
%MOIN*%
%OFA0B0*%
%SFA1.0B1.0*%
G01*
G36*
X564200Y442349D02*
X568100Y441149D01*
X574200Y437149D01*
X578400Y431049D01*
X579600Y426949D01*
X579900Y423049D01*
X579600Y419149D01*
X578400Y415149D01*
X574200Y409149D01*
X568200Y404949D01*
X564200Y403749D01*
X560300Y403349D01*
X556300Y403749D01*
X552300Y404949D01*
X546200Y409149D01*
X542200Y415149D01*
X541000Y419049D01*
X540600Y423049D01*
X541000Y427049D01*
X542200Y430949D01*
X546200Y437149D01*
X552400Y441149D01*
X556300Y442349D01*
X560300Y442749D01*
%LPC*%
G36*
X558300Y430649D02*
X557400Y430449D01*
X554700Y428649D01*
X552900Y425949D01*
X552700Y425249D01*
X552500Y423049D01*
X552700Y421249D01*
X553000Y420249D01*
X554700Y417649D01*
X557600Y415649D01*
X558300Y415449D01*
X560200Y415249D01*
X562300Y415449D01*
X562900Y415649D01*
X565700Y417649D01*
X567800Y420649D01*
X567900Y421249D01*
X568100Y423049D01*
X567900Y425249D01*
X567700Y425749D01*
X565700Y428649D01*
X563100Y430449D01*
X562300Y430649D01*
X560200Y430849D01*
%LPD*%
G36*
X201600Y459449D02*
X205500Y458249D01*
X211700Y454249D01*
X215600Y448149D01*
X217000Y444149D01*
[...]

PR #3370, opened 2017-12-03
close gerber regions properly (#3333)

AHorneffer:
The original code has only one "end region" (G37) statement for the whole file, but has multiple "start region" (G36) commands. AFAIK this results in an incorrect gerber file. (I think that polarity changes (%LPC or %LPD) inside a region are also illegal, but I'm too lazy to search the specifications for that.) 
This change puts the "end region" (G37) statement at the end of each region.


Intrinsically-Sublime:
Solution to https://github.com/fritzing/fritzing-app/issues/3333 NEW Gerber output broken.

Issue #3308, opened 2017-01-31
Parts Update Results In "Unable to open temporary file"


ghoss:
Version **0.9.3**, fresh install on **Mac OSX 10.9.5**

When I select _Help > Check for updates_, I get informed that a new parts library update is available. Fritzing then displays a progress window "Downloading new parts", and once finished, I get this error message:

    Sorry, unable to install new parts: Unable to open temporary file
    Fritzing must nevertheless be restarted, so the "Close" button will close Fritzing.

Upon restarting Fritzing, I get this error:

    Unable to find the following 13 parts:
    '4001_4_x_2_input_NOR_gate_single' at ''
    'JST-B3P-VH' at 'JST-B3P-VH.fzp'
    '4070_4_x_2_input_XOR_gate_multipart' at ''
    … (and so on)

The app will then continue to start up normally. If I select _Help > Check for updates_ again, this time I get the error:

    The parts folder '/Applications/Fritzing.app/Contents/MacOS/fritzing-parts' has been changed (4)
    Fritzing cannot update the parts folder automatically.
    If you used to git to make changes, use git to update and then regenerate the database by
    selecting Part > Regenerate parts database…
    If you did not use git to make changes, then you may need to reinstall Fritzing.

If I select _Part > Regenerate parts database_, I get the error

    Unable to open temporary file.

I also reinstalled the app (and deleted the preferences file `org.fritzing.Fritzing.plist` and the config data in `~/.config/Fritzing` before doing so). Same errors.

Any ideas?

tempidoro:
This is exactly also what I experienced. I need help

tempidoro:
I applied the suggestion given at
http://stackoverflow.com/questions/22661459/osx-with-xampp-imagegd2-unable-to-open-temporary-file
within another problem discussion.
it worked for me also in our case

ghoss:
@tempidoro Thanks for your help, and I can confirm that the [workaround at stackoverflow.com](http://stackoverflow.com/questions/22661459/osx-with-xampp-imagegd2-unable-to-open-temporary-file) indeed solves the problem for me too.

In short, open a shell in Terminal and type the command

`chmod 777 ${TMPDIR}`

Teddy68:
Hi all,

I had the same problem in Mountain Lion.
I first installed the Fritzing application in another folder than "Applications", because already over-crowded. At this point, I found no way to fix the problem.

After reading this : 
[...]

PR #3381, opened 2017-12-17
Bugfix for macOS in parts db regeneration

dlemper:
For details, have a look at this: https://github.com/fritzing/fritzing-app/issues/3308#issuecomment-351642253

### DIFF
--- a/src/fapplication.cpp
+++ b/src/fapplication.cpp
@@ -88,6 +88,7 @@ $Date: 2013-04-19 12:51:22 +0200 (Fr, 19. Apr 2013) $
 #include <QNetworkRequest>
 #include <QMultiHash>
 #include <QTemporaryFile>
+#include <QDir>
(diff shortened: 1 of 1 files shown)


tempidoro:
I apologize for the following request.
I'm not expert about both github and Fritzing. 
I have the same problem as described in #3308.
I understund that it has been solved by aknoerig making fritzing:master. 
However I do not know how to use the merged fritzing:master to update fritzing parts.
Thank you for any answer

dlemper:
Well, you could compile Fritzing from source (master branch) to get the already merged version..
Or you can wait for the next official release of fritzing, whenever that will come. Maybe @aknoerig knows, when that will be.
If you don't want to do that either, you can start your already installed version of Fritzing using the Terminal application. Then you can update your parts, too.

Hope this helps

Issue #3382, opened 2017-12-22
Trying to load a corrupted part hangs Fritzing.


vanepp:
Reproduce:

unzip a fzpz file and overwrite the fzp file in it with a copy of the fzpz file to make an invalid fzp file.

rezip the part to an fzpz file.

back up the user directories

F:\Fritzing\My Documents\fritzing\
F:\Fritzing\AppData\Roaming\fritzing

(using the names appropriate for your system) as we will corrupt them in the next step and you may lose
all your parts and sketches. 

start Fritzing 

file-open->corrupted.fzpz 

gets message box 

Parse error (2) at line 1, column1:
error occurred while parsing element 

F:\Fritzing\My Documents\fritzing\parts\user\part.prefix0000_d290581c8ba0365c66bddd27b5ec9988_1.fzp

click Ok to that message box and Fritzing seg faults and hangs. 

Restarting Fritzing after the hang gets the same error message as above (because it appears to have started to register the part before dying) until you clear the user directories. 

Fix:

At line 1982 in mainwindow.cpp it has returned from the part load but with a null in variable np (which should
point to the part) and attempting to use that causes the segfault and the hang. The solution would be to test
np for a NULL and if so go back out the changes already made (which I think are recorded somewhere) and return instead of trying to use the non loaded part. Easy to say, but I don't at present know how to do it. 
    I intend on poking at this more when I get some time but if someone else already knows what needs to be done to fix it so much the better. ...

PR #3388, opened 2018-01-08
fix bad part hangs Fritzing bug

vanepp:
This code seems to fix issue #3382 where a corrupted fzp file causes a seg fault in Fritzing and leaves the parts files around after the crash. While I don't know what I'm doing and just copied a section of code that deals with a similar situation and it would be good to have someone familiar with the code review this, since the alternative is a seg fault and a hang, I don't see that this fix could be worse.

Peter

### DIFF
--- a/src/mainwindow/mainwindow.cpp
+++ b/src/mainwindow/mainwindow.cpp
@@ -1900,7 +1900,17 @@ ModelPart* MainWindow::copyToPartsFolder(const QFileInfo& file, bool addToAlien,
 		}
 	}
 	ModelPart *mp = m_referenceModel->loadPart(destFilePath, true);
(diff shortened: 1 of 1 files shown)


KjellMorgenstern:
Will this change notice the user about the corrupted part? While it is desirable not to crash, leaving an corrupted part unnoticed might be even worse. What might have caused the corruption in the first place?

vanepp:
Yes it would alert the user that the part is corrupted. However since development has died, at this point it is moot unless we can restart development Iwhich some of us are trying to do).The corruption is caused by creating improperly formatted fzp or svg files (which is easy to do by accident). As you will likely see this pull request has not been acted on as development (and thus releases) have stopped.

Peter

PR #3389, opened 2018-01-21
Improve install script.

johndah4x0r:
I just added installation modes (user & system mode), since it will be handy to have it installled to the system. (The pre-made Ubuntu/Linux Mint package lacked the Fritzing icon) And I also removed 'sh' compatiblity. (C'mon! BASH 4 life!)


aknoerig:
Nice work, thanks!

PR #3391, opened 2018-01-26
"Further" improvements to both scripts.

johndah4x0r:
I tried to get finished with the uninstall script the fastest I can.
I will test the mechanism, and I will let you know if something has to be changed.


johndah4x0r:
And I will definitely "sacrifice" my system for a project THIS BIG!

PR #3401, opened 2018-03-16
Tent Vias

Intrinsically-Sublime:
Tent all vias removing them from the mask layers and allowing the silkscreen to go over the vias. This decreases the chances of possible shorts and prevents corrosion from flux trapped in the uncovered vias. It also looks better and allows the silkscreen to be more easily read.


KjellMorgenstern:
Would it be possible to provide screenshots of the desired effect?

deshipu:
I have been doing tented vias in Fritzing by generating the gerbers, then removing all vias from the design (select all vias, delete), and then generating gerbers again, and copying the silkscreen and soldermask layers from the second set of of gerbers into the first. The effect looks like this:

a72f12c7745a0adcac9040530e9365c9

benreu:
Is tented vias going to be mandatory, or is it possible to turn the tenting on/off per via? Some PCBs I would want this, but the next PCB it would be nice to have them untented, think about:

1 Using solder filled vias for heat transfer (some smd leds/mosfets use this trick)
2 Use solder filled vias to create a high current path from one layer to the next

maclof:
This should definitely be optional, on a per via basis. I would not want all my vias to be tented.

KjellMorgenstern:
I agree with @maclof that this should be optional. I think it could be made a property of the via. Workflows would be:

The inspector shows the propery "tented" . Additional if someone chooses  "select all vias", or selects multiple vias, the property should be applied to the selection (but that might be an additional fix, needed for Inspector settings in general)

PR #3448, opened 2019-05-01
fix Windows CI

ovidiub13:


KjellMorgenstern:
Although not yet optimized (no caching, dynamic linking is not ideal but no regression) , I'd like to merge it to develop now. It think it basically works.

PR #3455, opened 2019-05-04
Whitespace format Plist file

ovidiub13:


ovidiub13:
> The overall most used style on the fritzing-app is tabs, not spaces. We should not change this lighthanded, since whitespace issues cause a lot of conflict when merging. One exception could be if Apple discourages the use of tab in .plist files.

My personal preference is to use 4 spaces. But I have no problem aligning to what ever guideline a project has. Let's make a coding guidelines in the wiki, and show it in future PR a s checklist item.

This file was a mix of tabs and spaces. I aligned it all to 4 spaces. Is there an official document from Apple regarding `.plist` files syntax?

PR #3460, opened 2019-05-07
Add debugging target for the release script

KjellMorgenstern:


tools/linux_release_script/release.sh
```diff
@@ -8,18 +8,35 @@ failure() {
...
+else
+  relname=$1  #`date +%Y.%m.%d`
 fi
 
-#TODO Add debug-parameter to temporarily switch this of.
-if [[ -n $(git clean -xdn) ]]; then
-  echo "Build directory is not clean. Check git clean -xdn."
-  git clean -xfn
-  exit -1
+echo ${relname}
```
ovidiub13:
```suggestion
echo "Building release: ${relname}..."
```

tools/linux_release_script/release.sh
```diff
@@ -8,18 +8,35 @@ failure() {
...
+
+if [[ ${relname} == *"debug"* ]]; then
+  echo "Building a debug release"
```
ovidiub13: Not needed with the above proposed change.

PR #3467, opened 2019-05-11
Part handling

KjellMorgenstern:
This should fix a few errors that might appear when loading parts:
1. Relative paths specified on CLI caused problems
2. Recover from zero sized parts.db
3. Recover from missing parts.db
4. Avoid leaving zero sized parts.db, created valid parts.db when recovering.


d782:
parts.db zero size on ubuntu. :'C

d782:
Updated, i solved the Error "Sorry, we have a problem with the swapping mechanism. Fritzing still works, but you won't be able to change parts properties." 
apparently it's caused by a 0 byte parts.db file.
the solution is deleting, parts.db from /usr/share/fritzing/parts/    parts.db 
then Fritzing will do starting normally, i don't know exactly what's the error cause.

PR #3468, opened 2019-05-11
Support using system QuaZIP on Qt5 and output build messages for system libgit2 and QuaZIP

cjmayo:
* Currently Qt5 builds can't use system QuaZIP because it is using quazip5 locations.
* Add build-time messages to inform about the use of system libraries
* Don't output three copies of the build-time messages

A bit more info in the commit messages.

This is an updated version of the work in #3254.

Happy to squash commits in this PR or split into separate PRs if preferred.


cjmayo:
There may be more duplicate messages, suggested by old commit 41a3a04d170169c2caaad47780f6ae774115625d
this PR is dealing with the ones I see now on Linux with system Boost, libgit2 and QuaZIP.

phoenix.pro
```diff
@@ -213,8 +213,13 @@ include(pri/program.pri)
 include(pri/qtsysteminfo.pri)
 
 contains(DEFINES, QUAZIP_INSTALLED) {
-    INCLUDEPATH += /usr/include/quazip
-    LIBS += -lquazip
+    greaterThan(QT_MAJOR_VERSION, 4):exists(/usr/include/quazip5) {
+        INCLUDEPATH += /usr/include/quazip5
```
KjellMorgenstern: Qt4 is unsupported since 2015... we can not support it anymore, in fact we have to hurry so Fritzing can be ready for Qt6 by next year. So no such switches needed.

KjellMorgenstern:
I based the PR to fix-windows-ci, to make sure there are no conflicts.

phoenix.pro
cjmayo: Removed Qt 4 code.

cjmayo:
> I based the PR to fix-windows-ci, to make sure there are no conflicts.

I rebased the branch again on fix-windows-ci, hope that worked.

cjmayo:
Can't see why this was closed?

ovidiub13:
I guess so. Sorry for that.

PR #3469, opened 2019-05-12
Remove Qt 4 serialport support from phoenix.pro

cjmayo:
```diff
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -165,12 +165,7 @@ macx {
     QMAKE_BUNDLE_DATA += FILE_ICONS
 }
 
-QT += core gui svg xml network sql # opengl
-greaterThan(QT_MAJOR_VERSION, 4) {
-    QT += widgets printsupport concurrent serialport
-} else {
-    include($$QTSERIALPORT_PROJECT_ROOT/src/serialport/qt4support/serialport.prf)
-}
+QT += concurrent core gui network printsupport serialport sql svg widgets xml
 
 RC_FILE = fritzing.rc
 RESOURCES += phoenixresources.qrc
```

### DIFF
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -24,6 +24,10 @@
 #
 #********************************************************************/
 
+lessThan(QT_MAJOR_VERSION, 5) {
+    error(Fritzing does not build with Qt 4 or earlier)
+}
+
 # Fritzing requires two Qt-provided plugins in order to run correctly,
 # however the QTPLUGIN syntax only seems to work if Qt is built statically,
 # so QTPLUGIN is included here only for information purposes:
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -168,12 +168,7 @@ macx {
     QMAKE_BUNDLE_DATA += FILE_ICONS
 }
 
-QT += core gui svg xml network sql # opengl
-greaterThan(QT_MAJOR_VERSION, 4) {
-    QT += widgets printsupport concurrent serialport
-} else {
-    include($$QTSERIALPORT_PROJECT_ROOT/src/serialport/qt4support/serialport.prf)
-}
+QT += concurrent core gui network printsupport serialport sql svg widgets xml
 
 RC_FILE = fritzing.rc
 RESOURCES += phoenixresources.qrc


KjellMorgenstern:
Could you add a check, so it immediately fails when trying to build with qt4?

PS. I plan to create a development release (and hopefully 0.9.4) soon. So removing qt4 is currently not possible, as it would be to big of a change. So this PR would have to wait until 'develop' was merged to 'master'.

cjmayo:
Check added.

cjmayo:
Probably that should be a check for < Qt 5.6 because of:
```cpp
@@ -129,9 +114,7 @@ int main(int argc, char *argv[])
	int result = 0;
	try {
		//QApplication::setGraphicsSystem("raster");
#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0)
		QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
#endif
		FApplication * app = new FApplication(argc, argv);
```
https://github.com/fritzing/fritzing-app/pull/3470/files#diff-7ec3c68a81efff79b6ca22ac1f1eabba

KjellMorgenstern:
Indeed, I think support up to  < Qt 5.6 can be dropped.

cjmayo:
Not so simple. Ubuntu Xenial (16.04) has Qt 5.5 and Travis only got Xenial in November and doesn't look like getting Bionic any time soon:
https://travis-ci.community/t/ubuntu-18-04-1-lts-bionic-beaver/1270/

It is possible to get Travis to install a later Qt (which version?) from a ppa, this seems popular but I haven't done a lot of research:
https://launchpad.net/~beineri

Else I can leave this check at 5 and add the #if for Qt 5.6 back in #3470?

KjellMorgenstern:
Have a look at the travis files, on the develop branch, no decency problems
there, we can even run bionic. New distributions that include Fritzing
wouldn’t have a problem as well.

That said, thanks, 16.04 is a concern, we might want to keep it in

Chris Mayo <notifications@github.com> schrieb am Mo. 13. Mai 2019 um 20:59:

KjellMorgenstern:
Have a look at the travis files, on the develop branch, no decency problems
there, we can even run bionic. New distributions that include Fritzing
wouldn’t have a problem as well.

<del>That said, thanks, 16.04 is a concern, we might want to keep it in</del>

Sorry for the confusion. I already talked about this with Ovi... we'd rather put some effort to support AppImages or a similar technique instead of supporting old distros.

cjmayo:
Ah, I see it's using docker. I think it would be best if you could update the image to Bionic (Qt 5.9), adding a PPA will complicate .travis.yml and slow each CI run.

I've changed the check to be for at least Qt 5.6.

ovidiub13:
Let's keep this simple and just drop support for Qt 4 for now. We can handle <5.6 at a later time.
We're planning to get to use a package manager at some point (after the release) to handle dependencies.

Regarding the distribution of Fritzing on the various distros, we should have a proper documented discussion, in an Issue, not in a PR that seems to extend way beyond it's initial scope.

cjmayo:
OK. Removed the version check commit.

ovidiub13:
> Could you add a check, so it immediately fails when trying to build with qt4?

@cjmayo please add the ^ check.

cjmayo:
Added Qt version check back.

KjellMorgenstern (approved):
Looks good now, sorry for the confusion :-)

PR #3470, opened 2019-05-12
Remove compile-time checks for Qt 5.0

cjmayo:
Clearly this will have to wait too!

Translations will need updating for PostScript too:
```xml
    <message>
        <location filename="../src/mainwindow/mainwindow_export.cpp" line="146"/>
        <source>PostScript (*.ps)</source>
        <translation>PostScript (*.ps)</translation>
    </message>
```

### DIFF
--- a/src/dialogs/quotedialog.cpp
+++ b/src/dialogs/quotedialog.cpp
@@ -190,13 +190,8 @@ QuoteDialog::QuoteDialog(bool full, QWidget *parent) : QDialog(parent)
     m_tableWidget->verticalHeader()->setVisible(false);
     m_tableWidget->horizontalHeader()->setVisible(false);
     
-#if QT_VERSION < QT_VERSION_CHECK(5, 0, 0)
-    m_tableWidget->verticalHeader()->setResizeMode(QHeaderView::ResizeToContents);
-    m_tableWidget->horizontalHeader()->setResizeMode(QHeaderView::ResizeToContents);
-#else
     m_tableWidget->verticalHeader()->setSectionResizeMode(QHeaderView::ResizeToContents);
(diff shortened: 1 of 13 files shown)


cjmayo:
A lot shows up when I do `lupdate -no-obsolete phoenix.pro`, so I think translation updates need to be done carefully by someone looking at the translations themselves.

Added back the Qt 5.6 check so that this will pass Linux tests, and rebased on develop.

madpipeline (approved):
LGTM

ovidiub13:
I don't see any translation changes or the Qt 5.6 check you mention.

cjmayo:
> I don't see any translation changes

I ran the command to pick up on the removed translated strings:
`fileExtFormats[psActionType] = tr("PostScript (*.ps)");`
`m_exportPsAct = new QAction(tr("PostScript..."), this);`
which it did but it made all sorts of unrelated changes too. Therefore I didn't think it was safe.

> or the Qt 5.6 check you mention.

https://github.com/fritzing/fritzing-app/blob/1f88cce14a7f3a568c540ed45d5b5d91994e454d/src/main.cpp#L132

It was removed in the initial version of the PR but you didn't want to move to Xenial so I restored that change to get the PR to pass the CI check.

KjellMorgenstern:
Ah, I think that’s my misunderstanding here, there are some 5.6 checks in the code, but we just removed qt < 5 support, so there should be an explicit check in the build system for >= 5. So that you get a clean error message, not some compiler complains, if you try with an older qt version.

cjmayo:
I think the changes related to macOS were wrong, split them out into their own commit for better review.

madpipeline (approved):
LGTM. But I'll let @KjellMorgenstern merge.

KjellMorgenstern (approved):

PR #3472, opened 2019-05-14
Translations

KjellMorgenstern:
Updated the translations with qt lupdate,
resolved some conflicts, reuse some translation with minor changes in the English version, restructed the translator credits

### DIFF
--- a/src/help/aboutbox.cpp
+++ b/src/help/aboutbox.cpp
@@ -139,23 +139,24 @@ QString("<br /><br /><br /><br /><br /><br /><br /><br /><br />") +
 "</p>" +
 
 "<p>" +
-	tr("Thanks to Kurt Badelt and Miguel Solis for the Spanish translation, ") +
-	tr("to Gianluca Urgese for the Italian translation, ") +
-	tr("to Nuno Pessanha Santos for the Portuguese (European) translation, ") +
-	tr("to Yuelin and Ninjia  for the Chinese (Simplified) translation, ") +
-	tr("to Hiroshi Suzuki for the Japanese translation, ") +
-	tr("to Robert Lee for the Chinese (Traditional) translation, ") +
-	tr("to Vladimir Savinov for the Russian translation, " ) +
-	tr("to Steven Noppe and Davy Uittenbogerd for the Dutch translation, " ) +
-	tr("to Josef Dustira for the Czech translation, " ) +
-	tr("to Jinbuhm Kim for the Korean translation, " ) +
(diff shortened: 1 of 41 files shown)


src/help/aboutbox.cpp
```diff
@@ -139,23 +139,24 @@ QString("<br /><br /><br /><br /><br /><br /><br /><br /><br />") +
...
+	tr("Japanese: ") + tr("Hiroshi Suzuki ") +
+	tr("Russian: ") + "Vladimir Savinov " +
+	tr("Dutch: ") + "Steven Noppe, Davy Uittenbogerd " +
+	tr("Czech: ") + "Josef Dustira " +
+	tr("Korean: ") + tr("Jinbuhm Kim ") +
+	tr("Slovak: ") + " &#313;ubom&iacute;r Ducho&#328; " +
+    tr("Greek: ") + "Alexander Kaltsas " +
+    tr("French: ") + "Lionel Michel, Yvan Kloster, Alexandre Dussart, Roald Baudoux " +
+    tr("Turkish: ") + "Cihan Mete Bahad&#x0131;r " +
+    tr("Bulgarian: ") + "Nikolay Stankov, Lyubomir Vasilev " +
```
ovidiub13: Keep the same indentation for all lines please. Also, sorting the languages alphabetically would be nice.

ovidiub13: maybe also some vertical alignment, on columns, languages and names, it would be even clearer.

ovidiub13: Not all names are put in `tr()`, this breaks the translation files, most languages are missing the names.

Names shouldn't be translated IMO. So we should avoid using `and` or `&` in names strings, and just use commas.

We should not put names in `tr()`. Exception could be if traditionally names are rewritten in the non latin alphabets, Japanese, russian, greek, etc.

ovidiub13: And this is exactly what you did... Sorry, I just woke up. Perhaps do the same for Greek, Russian, Bulgarian, ?

KjellMorgenstern:
I put Asian names into tr because they often give themselves western names, that can be easily remembered and pronounced by foreigners.
Regarding the Unicode, this is a point, too. 

None of this yet is used in the names, the Chinese translators did not add their Chinese names, or they already got lost from the files in the past, so i‘d also be fine with YAGNI. Maybe it’s ok to use Unicode in the .cpp file?  I think all tools should support it (it’s supported with C++11)

PR #3475, opened 2019-05-16
Add --version command line option

cjmayo:
Split from #3242.

Copyright and mention of Friends-of-Fritzing e.V. need review.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -133,7 +133,9 @@ int main(int argc, char *argv[])
 		QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
 		FApplication * app = new FApplication(argc, argv);
-		if (app->init()) {
+		switch (app->init()) {
+        case FInitResultNormal:
+        {
 			//DebugDialog::setDebugLevel(DebugDialog::Error);
 			if (app->runAsService()) {
 				// for example: -g C:\Users\jonathan\fritzing2\fz\Test_multiple.fz -go C:\Users\jonathan\fritzing2\fz\gerber
@@ -149,8 +151,9 @@ int main(int argc, char *argv[])
 				}
 			}
 			app->finish();
-		}
-		else
+            break;
+        }
+		case FInitResultHelp:
         {
             QTextStream cout(stdout);
             cout <<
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -194,7 +197,23 @@ int main(int argc, char *argv[])
...
 				"The process path follows the -ep argument; the name of the menu item follows the -epname argument;\n"
 				"and any arguments to pass to the external process are provided in the -eparg arguments.\n";
+            break;
 		}
+        case FInitResultVersion:
+        {
+            QTextStream cout(stdout);
+            cout <<
+                "Fritzing " << Version::versionString() << "\n"
+                "Copyright (C) 2016 ???.\n"
```
ovidiub13: You can add `2019`. We plan several releases this year.

src/main.cpp
```diff
@@ -194,7 +197,23 @@ int main(int argc, char *argv[])
...
+                "Source code License GPLv3: GNU GPL version 3 <https://www.gnu.org/licenses/gpl.html>.\n"
+                "Documentation and Part Designs License CC BY-SA 3.0: Creative Commons Attribution-ShareAlike 3.0 Unported <https://creativecommons.org/licenses/by-sa/3.0/legalcode>.\n"
+                "This is free software: you are free to change and redistribute it.\n"
+                "There is NO WARRANTY, to the extent permitted by law.\n"
+                "\n"
+                "Maintained by the Friends-of-Fritzing e.V.\n";
```
ovidiub13: @KjellMorgenstern is this right?

src/main.cpp
```diff
@@ -194,7 +197,23 @@ int main(int argc, char *argv[])
...
+            break;
+        }
+        }
```
ovidiub13: Can you format the indentation?

madpipeline (approved):

src/main.cpp
ovidiub13: After some discussion with @KjellMorgenstern, we've think it's best for now to not have the © in the `--version` message due to the complicated legal situation of Fritzing right now.
In order not to hold this PR back: can you strip this message to just the version print? We'll revisit the © part later.

madpipeline (changes requested):

cjmayo:
Stripped back to version number only.
Replaced all tabs with spaces to improve the indentation.

KjellMorgenstern (approved):

PR #3478, opened 2019-05-17
Support using system QuaZIP on Qt5 and output build messages for system libgit2 and QuaZIP

cjmayo:
* Currently Qt5 builds can't use system QuaZIP because it is using quazip5 locations.
* Add build-time messages to inform about the use of system libraries
* Don't output three copies of the build-time messages

A bit more info in the commit messages.

This is an updated version of the work in #3254 and a copy of #3468.

Happy to squash commits in this PR or split into separate PRs if preferred.

### DIFF
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -198,7 +198,7 @@ include(pri/program.pri)
 include(pri/qtsysteminfo.pri)
 
 contains(DEFINES, QUAZIP_INSTALLED) {
-    message("using installed QuaZIP library")
+    !build_pass:message("using installed QuaZIP library")
     LIBS += -lquazip5
 } else {
     include(pri/quazip.pri)
@@ -207,4 +207,4 @@ contains(DEFINES, QUAZIP_INSTALLED) {
 TARGET = Fritzing
 TEMPLATE = app
 
-message("libs $$LIBS")
+!build_pass:message("libs $$LIBS")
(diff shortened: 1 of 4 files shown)


madpipeline (approved):
Looks ok to me. @KjellMorgenstern you ok if we merge this?

ovidiub13:
@cjmayo on what system and with what compiler are you building this?

cjmayo:
I'm using Gentoo Linux with GCC 8.3.0. QuaZIP 0.7.6.

KjellMorgenstern:
I see two issues with this PR.
1. Linking with QuaZIP works fine, why risk breaking things?
2. Linking against a system QuaZIP is not tested in CI.

I think the preferred way to fix this should be to remove the quazip code from Fritzing entirely, and adapt the CI accordingly. This way we would only have to maintain the dependency, as with every other library, not the library itself. Opinions?

KjellMorgenstern (changes requested):
I think it is better to remove the copy of quazip from Fritzing, and properly link against the library instead.

cjmayo:
Removing bundled QuaZIP sounds good to me.

cjmayo:
Updated onto latest develop. I don't use or know much about static linking but hopefully this could be the basis for still allowing dynamic linking to a system library and modified for static linking to a Fritizing copy.

PR #3479, opened 2019-05-18
Add build cache for Windows

ovidiub13:
Moving dependencies handling in dedicated script.
Add windows cache for Qt and LibGit2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,12 +53,12 @@ matrix:
 
   - os: windows
     language: cpp
+    cache:
+      directories:
+        - /c/Qt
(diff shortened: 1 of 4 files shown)


KjellMorgenstern:
It seems that the download of Qt is blocked, this is constantly failing. Works fine on my local machine, but travis can't access it... my guess is that they blocked travis for downloading it many times and not using a cache? Not sure wether travis has a cache proxy running, probably not.

ovidiub13:
The caching on Travis Windows does not work as advertised or expected, so closing this until we switch to using a package manager for dependencies.

PR #3482, opened 2019-05-23
Make pe delete part files on exit

vanepp:
This fixes issue 3383 where when a new part is created with parts editor, the filenames didn't get copied across to alien_files in the main thread. Since I'm new to both c++ and Qt someone experienced needs to look over the fix and make sure it is appropriate.

### DIFF
--- a/src/mainwindow/mainwindow.cpp
+++ b/src/mainwindow/mainwindow.cpp
@@ -2960,9 +2960,17 @@ QString MainWindow::getStyleSheetSuffix() {
     return "fritzing";
 }
 
-void MainWindow::addToMyParts(ModelPart * modelPart)
+void MainWindow::addToMyParts(ModelPart * modelPart, QStringList & m_peAlienFiles)
 {
-    if (modelPart != NULL) m_binManager->addToMyParts(modelPart);
+    if (modelPart != NULL) {
+        foreach(QString pathToAddFromPe, m_peAlienFiles) {
+            // DebugDialog::debug(QString("addToMyParts adding  %1")
+            //.arg(pathToAddFromPe));
+            m_alienFiles << pathToAddFromPe;
+            m_alienPartsMsg = tr("Do you want to keep the imported parts?");
+        }
+    }
+	m_binManager->addToMyParts(modelPart);
 }
 
 bool MainWindow::anyUsePart(const QString & moduleID) {
--- a/src/mainwindow/mainwindow.cpp
+++ b/src/mainwindow/mainwindow.cpp
@@ -2960,16 +2960,14 @@ QString MainWindow::getStyleSheetSuffix() {
     return "fritzing";
 }
 
-void MainWindow::addToMyParts(ModelPart * modelPart, QStringList & m_peAlienFiles)
+void MainWindow::addToMyParts(ModelPart * modelPart, QStringList & peAlienFiles)
 {
-    if (modelPart != NULL) {
-        foreach(QString pathToAddFromPe, m_peAlienFiles) {
+        foreach(QString pathToAddFromPe, peAlienFiles) {
             // DebugDialog::debug(QString("addToMyParts adding  %1")
             //.arg(pathToAddFromPe));
             m_alienFiles << pathToAddFromPe;
             m_alienPartsMsg = tr("Do you want to keep the imported parts?");
         }
-    }
 	m_binManager->addToMyParts(modelPart);
 }
 
--- a/src/mainwindow/mainwindow.cpp
+++ b/src/mainwindow/mainwindow.cpp
@@ -2960,7 +2960,7 @@ QString MainWindow::getStyleSheetSuffix() {
     return "fritzing";
 }
 
-void MainWindow::addToMyParts(ModelPart * modelPart, QStringList & peAlienFiles)
+void MainWindow::addToMyParts(ModelPart * modelPart, const QStringList & peAlienFiles)
 {
         foreach(QString pathToAddFromPe, peAlienFiles) {
             // DebugDialog::debug(QString("addToMyParts adding  %1")
--- a/src/mainwindow/mainwindow.h
+++ b/src/mainwindow/mainwindow.h
@@ -52,8 +52,6 @@ $Date: 2013-04-28 14:14:07 +0200 (So, 28. Apr 2013) $
 #include "../svg/svg2gerber.h"
 #include "../routingstatus.h"
 
-extern QStringList peAlienFiles;
-
 QT_BEGIN_NAMESPACE
 class QAction;
 class QListWidget;
--- a/src/mainwindow/mainwindow.h
+++ b/src/mainwindow/mainwindow.h
@@ -52,6 +52,8 @@ $Date: 2013-04-28 14:14:07 +0200 (So, 28. Apr 2013) $
 #include "../svg/svg2gerber.h"
 #include "../routingstatus.h"
 
+QStringList peAlienFiles;
+
 QT_BEGIN_NAMESPACE
 class QAction;
 class QListWidget;
@@ -456,7 +458,7 @@ protected slots:
 
     void keepMargins();
 	void dockChangeActivation(bool activate, QWidget * originator);
-    void addToMyParts(ModelPart *, QStringList &);
+    void addToMyParts(ModelPart *, const QStringList &);
     void hidePartSilkscreen();
     void fabQuote();
     void findPartInSketch();
@@ -908,7 +910,6 @@ protected:
     bool m_restarting;
 
     QStringList m_alienFiles;
-    QStringList peAlienFiles;
     QString m_alienPartsMsg;
     QStringList m_filesReplacedByAlienOnes;
 
--- a/src/mainwindow/mainwindow.h
+++ b/src/mainwindow/mainwindow.h
@@ -52,7 +52,7 @@ $Date: 2013-04-28 14:14:07 +0200 (So, 28. Apr 2013) $
 #include "../svg/svg2gerber.h"
 #include "../routingstatus.h"
 
-QStringList peAlienFiles;
+extern QStringList peAlienFiles;
 
 QT_BEGIN_NAMESPACE
 class QAction;
--- a/src/mainwindow/mainwindow.h
+++ b/src/mainwindow/mainwindow.h
@@ -908,7 +908,7 @@ protected:
     bool m_restarting;
 
     QStringList m_alienFiles;
-    QStringList m_peAlienFiles;
+    QStringList peAlienFiles;
     QString m_alienPartsMsg;
     QStringList m_filesReplacedByAlienOnes;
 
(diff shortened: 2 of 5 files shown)


src/mainwindow/mainwindow.cpp
```diff
@@ -2960,9 +2960,17 @@ QString MainWindow::getStyleSheetSuffix() {
     return "fritzing";
 }
 
-void MainWindow::addToMyParts(ModelPart * modelPart)
+void MainWindow::addToMyParts(ModelPart * modelPart, QStringList & m_peAlienFiles)
 {
```
KjellMorgenstern: The m_ prefix should be used for members, not parameters

KjellMorgenstern: I think, both modelPart and peAlienFiles, should be past as const reference.

src/mainwindow/mainwindow.cpp
```diff
@@ -2960,9 +2960,17 @@ QString MainWindow::getStyleSheetSuffix() {
...
+    if (modelPart != NULL) {
+        foreach(QString pathToAddFromPe, m_peAlienFiles) {
+            // DebugDialog::debug(QString("addToMyParts adding  %1")
+            //.arg(pathToAddFromPe));
+            m_alienFiles << pathToAddFromPe;
+            m_alienPartsMsg = tr("Do you want to keep the imported parts?");
+        }
+    }
+	m_binManager->addToMyParts(modelPart);
 }
```
KjellMorgenstern: The call to m_binManger now is outside of the null pointer check, was this intended?
This would be avoided by using const references (no null pointer check required, the method should not be called if it is expected to do nothing)

KjellMorgenstern:
Bump :-)

src/mainwindow/mainwindow.cpp
vanepp: @KjellMorgenstern  Sorry for the delay, I've been poking at this off and on. As I'm utterly new to c++ (I know c but not c++) that is taking some time. If I understand correctly, the m_peAlienFiles should change to just peAlienFiles (although I don't necessarily know why :-) ). The null pointer check is as intended (so that my code didn't do anything if there wasn't a modelPart present which I thought may occur under some case I don't understand), but from what you are saying should probably just be deleted if it won't get here without a modelPart. The other part I don't know about is if  the m_alienPartsMsg is correct. The data gets passed via the Qt IPC mechanism, but I'm not sure it is defined correctly (although it works and I don't see another way to get the definition from a single place in to both objects, there may be one). Once I figure out what needs to change the next challenge is going to be figuring out how to change the pull request (delete this one and submit a corrected one?) as I'm also new to (and usually baffled by) github.

src/mainwindow/mainwindow.cpp
KjellMorgenstern: > The m_ prefix should be used for members, not parameters

No problem, your help is welcome.
The name change is about 'coding style'.. rule of thump is that the m_ prefix is only for variables that are members. This way, when reading the code, one can see if changing the variable only has a local effect, or affects the whole instance.

src/mainwindow/mainwindow.cpp
KjellMorgenstern: The pull request will pick up change automatically. Just create another commit and push it to your branch (3383_fix...)

It looks to me that the null pointer will only occur when out of memory, so it is not as relevant. In fact, it might sense to omit it and just crash with a proper error message, but that would be part of a refactoring

src/mainwindow/mainwindow.h
```diff
@@ -908,7 +908,7 @@ protected slots:
     bool m_restarting;
 
     QStringList m_alienFiles;
-    QStringList m_peAlienFiles;
+    QStringList peAlienFiles;
```
KjellMorgenstern: Uhm... this is not the same variable... here it *is* a member variable of the class MainWindow, so the m_ prefix should be there.

src/mainwindow/mainwindow.cpp
```diff
@@ -2960,16 +2960,14 @@ QString MainWindow::getStyleSheetSuffix() {
     return "fritzing";
 }
 
-void MainWindow::addToMyParts(ModelPart * modelPart, QStringList & m_peAlienFiles)
+void MainWindow::addToMyParts(ModelPart * modelPart, QStringList & peAlienFiles)
 {
```
KjellMorgenstern: Here, peAlienFiles is a paramater to the method addToMyParts... its scope is limited to the method. So it correctly does not have the m_ prefix. However, since peAlienFiles is not modified, it should be const.

src/mainwindow/mainwindow.cpp
```diff
@@ -2960,9 +2960,15 @@ QString MainWindow::getStyleSheetSuffix() {
     return "fritzing";
 }
 
-void MainWindow::addToMyParts(ModelPart * modelPart)
+void MainWindow::addToMyParts(ModelPart * modelPart, QStringList & peAlienFiles)
```
KjellMorgenstern:
```suggestion
void MainWindow::addToMyParts(ModelPart * modelPart, const QStringList & peAlienFiles)
```

src/mainwindow/mainwindow.h
```diff
@@ -908,6 +908,7 @@ protected slots:
     bool m_restarting;
 
     QStringList m_alienFiles;
+    QStringList peAlienFiles;
```
KjellMorgenstern:
```suggestion
    QStringList m_peAlienFiles;
```

src/mainwindow/mainwindow.h
```diff
@@ -456,7 +456,7 @@ protected slots:
 
     void keepMargins();
 	void dockChangeActivation(bool activate, QWidget * originator);
-    void addToMyParts(ModelPart *);
+    void addToMyParts(ModelPart *, QStringList &);
```
KjellMorgenstern:
```suggestion
    void addToMyParts(ModelPart *, const QStringList &);
```

src/partseditor/pemainwindow.h
```diff
@@ -83,7 +83,7 @@ Q_OBJECT
     void changeOriginalFile(ViewLayer::ViewID viewID, const QString originalFile, int changeCount);
 
 signals:
-    void addToMyPartsSignal(ModelPart *);
+    void addToMyPartsSignal(ModelPart *, QStringList &);
```
KjellMorgenstern:
```suggestion
    void addToMyPartsSignal(ModelPart *, const QStringList &);
```

vanepp:
@KjellMorgenstern Ok, I have removed the NULL check, and changed m_peAlienFiles to peAlienFiles, compiled and tested that the fix still works fine locally and pushed the commit (github is slick when I don't screw up and break something :-) )  so hopefully we should be good to go. 

Peter

src/mainwindow/mainwindow.h
vanepp: That should mean this declaration is in the wrong place then. What is happening is the parts editor is running in one thread and creates the parts files, the file names of which need to go in to m_alienFiles in the mainwindow thread so it knows to delete them if the user chooses to (the parts editor does not delete them because that is user choice later). Here that is achieved by a Qt ipc call (signal) to basically copy a local variable from one process to the other (which seems to me to be a violation of the OOP principles). So I appear to either have the define in src/mainwindow/mainwindow.h in the wrong place, or this is not the correct way to do what I want to do (although I think I copied this from something else that did much the same thing, but that is a year and a half ago at this point). The idea is the src/mainwindow/mainwindow.h file is included in both mainwindow.cpp and in pemainwindow.cpp so they both know what how the  local variable that has appeared via IPC is defined. It seemed to me it doesn't want to be public, so protected (while I don't really know what that means, I had assumed private to the module) the same as the current m_alienFiles seemed the best bet and the compiler appears happy with it as the code appears to work. Is there some other way to define peAlienFiles so it is defined in both pemainwindow.cpp and mainwindow.cpp, or perhaps is the correct answer to not do the IPC call at all but copy the file names in to m_alienFiles directly which would mean getting m_alienFiles passed in to pemainwindow.cpp from somewhere . I think the file names are destroyed just after the current signal call and the part is passed as a model structure without the required file names to the mainwindow function. This method works, but that doesn't mean it is the best way to do this.

KjellMorgenstern: From a modeling point of view yes, the main window should not need to know details about parts, for example that parts include so alienFiles. 
How this is best realized with the existing structure I can not tell. Just the fact that a protected or private variable is modified via an IPC call is not a violation of OOP. But bloating implementation details into one giant class is.
There is already a lot of stuff in mainwindow (the sheer file size and number of members is quite a strong indicator that there is something wrong). But is it possible handle this in the parts editor window?

vanepp: "There is already a lot of stuff in mainwindow (the sheer file size and number of members is quite a strong indicator that there is something wrong). But is it possible handle this in the parts editor window?"

Unfortunately  no, while pemainwindow creates the files, it then puts the part in the mine parts bin without the file names being added to alienFiles (that may be where this should be done but I don't know how yet) . In mainwindow.cpp as Fritzing is exiting, it asks if you want to save the parts in the mine parts bin. If the user says no, then it uses the file names in alienFiles to delete the no longer needed files from the parts directory. Currently, (and what this fix tries to correct) the files for the part created by parts editor aren't in alienFiles and thus don't get deleted, which means you can'r reload that same part again because the files are present, but you can't delete the files from within Fritzing because the part no longer exists (only the files) and you need to manually remove the files from the parts directory outside of Fritzing. I made a push with the requested changes and an attempt at fixing this problem by moving the peAlienFiles definition to the top of the mainwindow.h file, but that doesn't compile as the symbol is multiply defined. I expect it needs to be extern in the .h file, and defined in one or another of the .cpp files but I'm having trouble figuring out which one it should be in (I think pemainwindow.cpp). I now have a version that compiles, but it doesn't seem to run correctly. I'll keep poking.

vanepp:
With the latest commit I think it may be done (assuming you don't want to scrap this fix and do it another way). My main problem turned out to be forgetting there is a signal definition in src/mainwindow/mainwindow_menu.cpp which hadn't changed for the const added to the arguments. Updating that so the signal actually gets through makes the fix work again.

src/mainwindow/mainwindow.h
```diff
@@ -52,6 +52,8 @@ along with Fritzing.  If not, see <http://www.gnu.org/licenses/>.
 #include "../svg/svg2gerber.h"
 #include "../routingstatus.h"
 
+extern QStringList peAlienFiles;
+
```
KjellMorgenstern: I think this isn't used anymore, you can remove the line?

vanepp:
You are correct, I assume it is figuring out that peAlienFiles should be a QStringList from the parameter definitions (that is why I added the extern). The fix compiles and works without that define so all should no  be well. 

Peter

KjellMorgenstern:
Thanks a lot :-)

PR #3487, opened 2019-06-19
Develop

KjellMorgenstern:
Add a continuous delivery configuration:
For each commit to the develop branch,  try to create binaries for Linux and Mac and Windows.
If this is successful, the Linux and Mac build will be published via github releases. 

This works the following way:
Via travis stages, the individual jobs are coordinated.
Successful builds are collected are copied from the travis jobs container to AWS S3
If everything worked, builds are then copied from S3 to github releases, and the release is tagged with the build number.


KjellMorgenstern:
PR now managed on fritzing organization ( PR #3488 )

PR #3497, opened 2019-07-04
Arduino no ftdi fzz fix

vanepp:
Fix Arduino-no-FTDI.fzz in examples. Currently C9 is backwards (- to +5V + to ground). So reverse C9 in all three views. While I'm here some cosmetic changes in schematic as well: change yellow wires on crystal and D5 to orange to make clicking on a connection (which turns the connected pads yellow) more visible. Align the capacitors for neatness. Make the wire on the top of the crystal level rather than slanted for neatness.


KjellMorgenstern:
I am sure this caused some smoking capacitors, thanks for fixing! Check with a recent Fritzing build looks good. If you don't have have further changes, I'll manually push #2e79d21 directly to develop, since there are some merge errors.

vanepp:
I don't have more changes, I suspect I'm at fault for the merge errors, when I did the pull it complained about my branch being out of date and offered to fix it (which I accepted). I don't know why my branch is out of date since I think I synced it lately.

PR #3502, opened 2019-07-08
Prettify repo, improve readability of build files

Phikimon:
As a person who had zero experience with QT and little experience with CI management and cross-platform builds, I can say that it takes a lot of time to find out what's actually going on in .{pro,pri} files, what 'config.tests' are etc. This pull-request is intended to ease suffering of newcomers. It mostly consists of refactoring, renames, comments adding etc.

Some of the commits might seem quite radical(see last two of them), so I would agree to drop them if @KjellMorgenstern thinks that's appropriate.

If this PR will be merged, changes to the fritzing-app/wiki are also desired.

Disclaimer: despite my best effort not to crash anything, changing of build files is quite risky, so please check the changes thoroughly.


Phikimon:
P.P.S. I managed to build Fritzing with this PR applied on my local Debian Stretch 4.9.0 and Mac OS 10.14.5, let's see what CI thinks about that :)

Phikimon:
Only changed git-commits metadata

KjellMorgenstern:
Qt 5.9 is to early, we still stick with 5.5 on linux (because of ubuntu 16.04).

Splitting the build system into windows / linux / mac version is quite the oposite of what I am trying to do. git bash can be used, and it seems windows is going to improve support for bashisms. So, future steps would be to refactor the .bat files into .sh. Some code (translations, fritzing-parts) can be unified.
Similar with the ci system. CI and Manual builds should behave as similar as possible. So I'd rather create separate directories for each of those. 

I like the way whitespaces have been fixed. Trailing whitespaces should be removed, tabs should be used or indentation, at least in C++. For bash or other scripts this may vary. 

It might make sense to split the PR, so the less controversial parts can be moved in.

KjellMorgenstern:
For the .pro and .pri files. We will have to move from qmake to cmake, so don't put to much time into that.

KjellMorgenstern:
About removing the compiled translations: Should be done, but release scripts for all platforms need to be verified to properly install translations tools and build the translations.

KjellMorgenstern:
About using pkg-config ... of course fixes to the current process are welcome. However when switching to cmake, pkg-config is not the default way to go (use find_package and cmake modules instead).

KjellMorgenstern:
I cherry picked some commits that should already be clear: https://github.com/fritzing/fritzing-app/pull/3503

KjellMorgenstern:
Created issues #3504 , #3505, #3506, #3507 about build system questions that concern this PR. Closing the PR for now. It is split up in PR #3503 and the above mentioned four issues.

Phikimon:
> tabs should be used or indentation, at least in C++
Are you sure? 
Quoute from https://wiki.qt.io/Qt_Coding_Style :
> 4 spaces are used for indentation
> Spaces, not tabs!

KjellMorgenstern:
Fritzing c++ codes uses tabs a lot, so I'd be willing to override the qt style in that point. I think IDEs that mess this up are luckily a problem of the past. 
We deviate from qt style in some cases. For example, not sure why qt allows (even suggests) "if"-branches without block delimiters. 
https://github.com/fritzing/fritzing-app/wiki/2.4-Coding-style

PR #3513, opened 2019-07-20
reinstate deleted Creator Kit jar files

cjmayo:
Files were deleted by commit:

d4abfc0c ("Fix hundreds of whitespace errors", 2019-07-07)

---

I haven't used these but N.B. they were deleted.


KjellMorgenstern:
Good spot, thank you. Anybody with some knowledge about the processing toolchain? Four copies of the same binary plus two copies of an older version in a git repo is certainly not ideal.

KjellMorgenstern:
Seems this is unsused.

PR #3515, opened 2019-07-21
Remove bundled QuaZIP

cjmayo:
This is #3478 reworked to only use the system QuaZIP as suggested in https://github.com/fritzing/fritzing-app/pull/3478#pullrequestreview-261543160.


cjmayo:
What needs to be done to get the quazip headers installed for testing?:

```
/home/conan/fritzing/src/utils/folderutils.cpp:36:10: fatal error: quazip5/quazip.h: No such file or directory

 #include <quazip5/quazip.h>
```
I've tried to include them:
https://github.com/fritzing/fritzing-app/pull/3515/commits/300a1f1e77837c14aa108ca8c4ee98925a5d6186
but does it mean the image needs recreating?

KjellMorgenstern:
Yes, the docker images would need to be build. In theory, it would also be possible to use docker compose, to add a separate libquazip-Docker container.
But the bigger problem is: By linking libquazip5 dynamically, every machine, needs to add this dependency. This is ok for a fritzing.deb or rpm package. But it is not ok for the main distribution we have : zip, dmg and tarballs.

KjellMorgenstern:
So, at least we need to make dynamic linking an option, and use static linking by default. I am not sure how WIndows and Mac handle libquazip. It would probably be best to pull the source code and build the library by ourselfes, similar to libgit2. Considering the effort for this, and the small size of the library, maybe it is better to just update the copy fritzing has?

cjmayo:
OK. I'll close this one then and update the original PR which is probably closer to that.

PR #3518, opened 2019-08-05
Add built target for fedora

KjellMorgenstern:


docker/Dockerfile.fedora
```diff
@@ -0,0 +1,23 @@
+FROM fedora:31
+
+ARG user_id=1000
+ENV env_user_id=${user_id}
+
+RUN groupadd -g ${env_user_id} conan && \
+    useradd -r -u ${env_user_id} -g conan conan
+
+USER root
```
ovidiub13: You are root at this point. There is no need for this change.

docker/Dockerfile.fedora
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /home/conan
```
ovidiub13: This directory needs to exist in order to change to it.

docker/Dockerfile.fedora
```diff
@@ -0,0 +1,23 @@
...
+
+RUN dnf -y install openssl-devel zlib-devel libjpeg-devel libpng-devel bzip2 libgit2-devel boost-devel
+RUN dnf -y install xorg-x11-server-Xvfb qt-creator qt5-devel
+
+RUN ln -s /usr/bin/qmake-qt5 /usr/bin/qmake
+RUN ln -s /usr/bin/lrelease-qt5 /usr/bin/lrelease
+
+RUN dnf clean all
+
+USER conan
```
ovidiub13: Please add trailing end line.

docker/Dockerfile.fedora
KjellMorgenstern: I added the -m option to the useradd command

PR #3539, opened 2019-10-06
Use Constructor Member Initialization syntax to improve performance

DrItanium:
I ran fritzing-app through the static analysis tool my company makes (we have a release coming up and I thought fritzing would be a perfect project to test on) and found a considerable number of places (I know there are more but I have to check the SA results again) where:

1. Members of a class are being initialized twice (once during object construction and then in the body of the constructor)
2. Not all members of a class are initialized. This can lead to undefined behavior
3. One instance of "C-ing" in C++. I believe the objective was to zero out memory on memory allocation, I have fixed that so malloc is not called and instead new. Adding parens after the square brackets causes the memory to be initialized to zero (or a constructor to be called, which I didn't know before researching this). I also eliminated the free call and integrated the logic into the destructor as well. 
4. Using forward decls to cut down on typing. 

Oh, for more information. When you use the constructor list initialization syntax, you are placing the values into the members as the object is being constructed prior to the body of the constructor being called. When you assign a value to a field within the body of the constructor, you are actually doing a second write to that member. PODs do not have a constructor so they must be explicitly initialized either via Non-Static Data Member Initialization (NSDMI) or through the initializer list syntax. The order of member init exactly matches the layout of the members in the class/struct so I had to rearrange in a few places. In the case of class instances, their default constructor is called implicitly so you technically don't need to call their default constructor explicitly but if you don't initialize in order you'll get a warning from the compiler about out of order member initialization which can be bad. 

There are cases where it makes more sense to assign to a member inside the body of the constructor so I left those where it makes sense. 

The other change I made is started using constexpr all the places I could. The downside to this approach is it breaks compatibility with non C++11 compilers. I may have even broken compatibility with C++11 and required C++14 (I'm not sure since C++14 is so useful I use features from it without even thinking). I bring this up because I see references to MSVC 2010 (an ancient but safe compiler target, I would actually recommend MSVC 2015 or later these days if you want/must build with MSVC on windows). 

I am not done fixing what we call severity one violations. But I thought it would be smart to not give too large of a pull request.

### DIFF
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -1365,10 +1365,10 @@ QString WireWidthChangeCommand::getParamString() const {
 ///////////////////////////////////////////
 
 RoutingStatusCommand::RoutingStatusCommand(SketchWidget * sketchWidget, const RoutingStatus & oldRoutingStatus, const RoutingStatus & newRoutingStatus, QUndoCommand * parent)
-	: BaseCommand(BaseCommand::SingleView, sketchWidget, parent)
+	: BaseCommand(BaseCommand::SingleView, sketchWidget, parent),
+	m_oldRoutingStatus(oldRoutingStatus),
+	m_newRoutingStatus(newRoutingStatus)
 {
-	m_oldRoutingStatus = oldRoutingStatus;
-	m_newRoutingStatus = newRoutingStatus;
 }
 
 void RoutingStatusCommand::undo() {
@@ -1393,11 +1393,11 @@ QString RoutingStatusCommand::getParamString() const {
 ///////////////////////////////////////////////
 
 ShowLabelFirstTimeCommand::ShowLabelFirstTimeCommand(SketchWidget *sketchWidget, CrossViewType crossView, long id, bool oldVis, bool newVis, QUndoCommand *parent)
-	: BaseCommand(crossView, sketchWidget, parent)
+	: BaseCommand(crossView, sketchWidget, parent),
+	m_itemID(id),
+	m_oldVis(oldVis),
+	m_newVis(newVis)
 {
-	m_itemID = id;
-	m_oldVis = oldVis;
-	m_newVis = newVis;
 }
 
 void ShowLabelFirstTimeCommand::undo()
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -53,10 +53,10 @@ CommandProgress BaseCommand::m_commandProgress;
 
 BaseCommand::BaseCommand(BaseCommand::CrossViewType crossViewType, SketchWidget* sketchWidget, QUndoCommand *parent)
 	: QUndoCommand(parent),
-    m_crossViewType(crossViewType),
-    m_sketchWidget(sketchWidget),
-    m_parentCommand(parent),
-    m_index(BaseCommand::nextIndex++)
+	m_crossViewType(crossViewType),
+	m_sketchWidget(sketchWidget),
+	m_parentCommand(parent),
+	m_index(BaseCommand::nextIndex++)
 {
 }
 
--- a/src/commands.cpp
+++ b/src/commands.cpp
@@ -1770,11 +1770,11 @@ QString TransformItemCommand::getParamString() const {
 
 SetResistanceCommand::SetResistanceCommand(SketchWidget * sketchWidget, long itemID, QString oldResistance, QString newResistance, QString oldPinSpacing, QString newPinSpacing, QUndoCommand * parent)
 	: BaseCommand(BaseCommand::CrossView, sketchWidget, parent),
-    m_oldResistance(oldResistance),
-    m_newResistance(newResistance),
-    m_oldPinSpacing(oldPinSpacing),
-    m_newPinSpacing(newPinSpacing),
-    m_itemID(itemID)
+	m_oldResistance(oldResistance),
+	m_newResistance(newResistance),
+	m_oldPinSpacing(oldPinSpacing),
+	m_newPinSpacing(newPinSpacing),
+	m_itemID(itemID)
 {
 }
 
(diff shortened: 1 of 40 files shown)


DrItanium:
Awesome :). It makes a lot more sense to wait until 0.95 for changes like
this. I'll revise this pull request against develop.

src/svg/gedaelementlexer.cpp
```diff
@@ -25,27 +25,32 @@ along with Fritzing.  If not, see <http://www.gnu.org/licenses/>.
...
+    m_comments()
 {
-	m_nonWhitespaceMatcher.setPattern("[^\\s]");
-	m_commentMatcher.setPattern("(^\\s*\\#)");
-	m_elementMatcher.setPattern("Element\\s*([\\(\\[])");
-	//m_stringMatcher.setPattern("\"([^\"]*)\"");
-	m_stringMatcher.setPattern("\"([^\"\\\\]*(\\\\.[^\"\\\\]*)*)\"");
-	m_integerMatcher.setPattern("[-+]?\\d+");
-	m_hexMatcher.setPattern("0[xX][0-9a-fA-F]+");
+    /// @todo figure out how to do inplace initialization of a cleaned source string.
```
KjellMorgenstern: The clean() method is quite expensive, compared to the QString copy/move operation, maybe this todo is a bit overkill?

src/commands.cpp
```diff
@@ -60,13 +52,12 @@ CommandProgress BaseCommand::m_commandProgress;
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 BaseCommand::BaseCommand(BaseCommand::CrossViewType crossViewType, SketchWidget* sketchWidget, QUndoCommand *parent)
-	: QUndoCommand(parent)
+	: QUndoCommand(parent),
+    m_crossViewType(crossViewType),
+    m_sketchWidget(sketchWidget),
+    m_parentCommand(parent),
+    m_index(BaseCommand::nextIndex++)
```
KjellMorgenstern: Here (and some other places) please use tabs for indentation.

DrItanium: I'll fix that, my vimrc is responsible for the mismatch, looks correct on my side but it actually isn't.

src/partsbinpalette/graphicsflowlayout.cpp
```diff
@@ -40,19 +40,18 @@ void GraphicsFlowLayout::setGeometry(const QRectF &rect) {
...
-	double y = rect.y();
-	double lineHeight = 0;
+    auto x = rect.x();
+    auto y = rect.y();
+    auto lineHeight = 0.0;
 
-	for(int i=0; i < count(); i++) {
+	for(int i=0; i < count(); ++i) {
 		QGraphicsLayoutItem* item = itemAt(i);
-		int nextX = x + item->preferredSize().width() + spacing();
```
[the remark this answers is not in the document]
DrItanium: Huh, I don't remember why I did it... But I always favor auto over an explicit type to prevent the issue of unintended type conversion.

src/svg/gedaelementlexer.cpp
DrItanium: With guaranteed copy elision in c++11/14, the value returned is actually the memory within the instance. You're right that clean is very expensive. That is exactly why I added the todo, I figured that it may be beneficial but I wasn't totally convinced. Just didn't want to forget about it.

src/connectors/bus.cpp
```diff
@@ -24,19 +24,23 @@ along with Fritzing.  If not, see <http://www.gnu.org/licenses/>.
...
+    m_busShared(busShared),
+    m_modelPart(modelPart)
 {
-	m_busShared = busShared;
-	m_modelPart = modelPart;
 }
 
-const QString & Bus::id() const {
-	if (m_busShared == NULL) return ___emptyString___;
-
```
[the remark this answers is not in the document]
DrItanium: I was getting massively confused on how this code was working so I used the syntax we use at work and my own projects. I have no issue conforming to your coding style. I would actually suggest using c++17's std::optional in the future instead of empty string. I'll fix that as well.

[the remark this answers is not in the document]
KjellMorgenstern: Probably we can not switch to C++17 that soon. But yes, there are to many nullptr checks in the Fritzing code, some of them looking quite defensive.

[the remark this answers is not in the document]
DrItanium: Using std::optional would require some heavy refactoring. I would not recommend it in the short term anyway. C++17 is wonderful but having access to C++14 is so welcome :).

DrItanium:
Have some down time at work waiting for tests to run. Making changes as described and also found several cases where I missed using list init syntax.

src/partsbinpalette/graphicsflowlayout.cpp
```diff
@@ -40,19 +41,18 @@ void GraphicsFlowLayout::setGeometry(const QRectF &rect) {
...
+	auto lineHeight = 0.0;
 
-	for(int i=0; i < count(); i++) {
+	for(int i=0; i < count(); ++i) {
 		QGraphicsLayoutItem* item = itemAt(i);
-		int nextX = x + item->preferredSize().width() + spacing();
+		auto nextX = x + item->preferredSize().width() + spacing();
 
 		if (item->sizePolicy().horizontalPolicy() == QSizePolicy::Expanding) {
 			int myY = y + lineHeight + spacing() + SpaceBefore;
```
KjellMorgenstern:
```suggestion
			auto myY = y + lineHeight + spacing() + SpaceBefore;
```

DrItanium:
I made the suggested change. What do you mean by QPoint?

PR #3541, opened 2019-10-09
Disable support for qt < 5.9

KjellMorgenstern:


src/main.cpp
```diff
@@ -55,21 +54,22 @@ void writeCrashMessage(const QString & msg) {
...
+		case QtWarningMsg:
+		originalMsgHandler(type, context, msg);
+			break;
+		case QtCriticalMsg:
+		originalMsgHandler(type, context, msg);
+			break;
+		case QtFatalMsg:
+			{
+				writeCrashMessage(msg);
+			}
```
ccoenen: This indentation looks really weird. Also, the last one has braces and the others do not? This looks like this might create problems down the road (see [2014's goto fail](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/) bug)

KjellMorgenstern: Indeed... i'd also argue it is a bad idea to catch a Fatal , log it, and then just hide it as a Warning.
How about replacing the whole thing with:

```suggestion
void fMessageHandler(QtMsgType type, const QMessageLogContext &context, const QString & msg)
{
	if (type == QtFatalMsg) {
		writeCrashMessage(msg);
	}
	originalMsgHandler(type, context, msg);	
}			
```

KjellMorgenstern:
I have integrated my suggested change: c52eb6a1 .

PR #3547, opened 2019-10-23
Update Russian translation

OlesyaGerasimenko:


translations/fritzing_ru.ts
```diff
@@ -2186,17 +2186,17 @@ It is also possible to choose a connector as a ground fill seed by right-clickin
...
         <location filename="../src/mainwindow/mainwindow.cpp" line="2003"/>
         <source>%1 - [%2]</source>
-        <translation>%1 - [%2]</translation>
+        <translation>%1 — [%2]</translation>
     </message>
     <message>
         <location filename="../src/mainwindow/mainwindow.cpp" line="1433"/>
         <source>No Sketch found in &apos;%1&apos;</source>
-        <translation>Не найдено эскизов в &apos;%1&apos;</translation>
+        <translation>Не найдено эскизов в «%1»</translation>
```
KjellMorgenstern: Please conserve special codes like `&apos;  `

translations/fritzing_ru.ts
```diff
@@ -2186,17 +2186,17 @@ It is also possible to choose a connector as a ground fill seed by right-clickin
     <message>
         <location filename="../src/mainwindow/mainwindow.cpp" line="2003"/>
         <source>%1 - [%2]</source>
-        <translation>%1 - [%2]</translation>
+        <translation>%1 — [%2]</translation>
```
KjellMorgenstern: Why did you replace - with — ?

translations/fritzing_ru.ts
```diff
@@ -1269,22 +1269,22 @@ Use the file?</source>
...
+        <translation>&lt;br/&gt;Режим&lt;b&gt;Представление макетной платы&lt;/b&gt; позволяет показать, как выглядит &lt;i&gt;настоящий&lt;/i&gt; макет.&lt;br/&gt;&lt;br/&gt;Начните с перетаскивания одного компонента из каталога компонентов, который находится в правом верхнем углу. Затем перетащите остальные компоненты, соединяя их путём размещения на макете или щёлкая по разъёмам и перетаскивая провода. Этот процесс похож на то, как вы располагали бы вещи в физическом мире. &lt;br/&gt;&lt;br/&gt;После завершения создания эскиза в режиме макета попробуйте использовать другие предоставления. Можно переключать представления, щёлкая по вкладкам в верхней части окна. Поскольку разные представления имеют разное назначение, компоненты будут выглядеть по-разному в различных видах.</translation>
```
KjellMorgenstern: Please conserve special codes like `&quot; `

OlesyaGerasimenko: OK, I'll fix this.

translations/fritzing_ru.ts
OlesyaGerasimenko: This is according to the rules of the Russian language («—» should be used between phrases; «-» cannot be preceded or followed by spaces).

translations/fritzing_ru.ts
OlesyaGerasimenko: OK, though according to the Russian language rules these quotes should be used: «»

translations/fritzing_ru.ts
OlesyaGerasimenko: (Though my version would be true to source formatting.)

KjellMorgenstern: Ah, using different signs would be ok. However I don't know how by heart how those are represented corretly in Qt. Do you have the possibility to generate a build and check if it looks correct?

translations/fritzing_ru.ts
KjellMorgenstern: It is not used for phrases, but for separating the filename and title in the Window title. 
I looked up a random screenshot to show what I mean, see the windows title here: http://s020.radikal.ru/i708/1702/d9/cce20896172b.png . Does this rule still apply?

OlesyaGerasimenko: Yes, it still applies in this case.

translations/fritzing_ru.ts
OlesyaGerasimenko: Unfortunately, I don't have the skills to make a build, but from my previous experience I can tell that the «» quotes do look okay in programs built with .ts (.qm) files containing them.

OlesyaGerasimenko: Checked the translation in the application (by copying the .qm file to usr/share) and made a few changes, so it looks fine now.
Screenshot_20191023_170231

OlesyaGerasimenko: The «» quotes look okay as well.
Screenshot_20191023_170716

PR #3549, opened 2019-10-23
Master

OlesyaGerasimenko:


KjellMorgenstern:
Please use the develop branch for pull requests. It can be easily moved to develop, because there are no other changes since the last sync of master and develop. But it would cause bigger issues later.

PR #3555, opened 2019-11-01
Fix #2186

### LINKED ISSUE #2186: ground fill is connected only to the lower side
_From [th.gauwe...@googlemail.com](https://code.google.com/u/114319454448349078647/) on September 03, 2012 05:52:49_

What steps will reproduce the problem? 1. define a pin hole
2. set ground fill seeds
3. fill up with gnd

The pins are only connected to GND on the lower side, but not to the upper side of the pcb.

The file contains the connector for an antenna at the bottom right.
The four pins around the middle pin are connected to GND.

**Attachment:** [RFM12 breakout masse.fzz Screenshot gc viewer antenna.png](http://code.google.com/p/fritzing/issues/detail?id=2186)

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=2186_

KjellMorgenstern:


KjellMorgenstern:
image

When connecting a seed element, the size of the seed element was used to check if there are any other items in the way. I corrected the code to only check for about the size of the small copper area that is added as a connector, plus about 18mil  ( 0.5mm) in each direction.

PR #3585, opened 2019-12-24
Simplify the != NULL checks

DrItanium:
I used sed to remove all the != NULL checks from the code. All of the == NULL checks are still there in this code review. Those are a little more involved to fix.


DrItanium:
Since I am already in there, I went and replaced all instances of NULL with nullptr as well.

DrItanium:
I'm pulling this review request as I broke something.

PR #3593, opened 2020-01-11
General cleanup of ItemBase and NonConnectorItem

DrItanium:
Standard nullptr check cleanup, NSDMI, and constexpr'ing all the things for NonConnectorItem and ItemBase.


DrItanium:
Did I break something?

KjellMorgenstern:
Probably not, the automated Windows build currently has an issue.

PR #3618, opened 2020-02-13
Allow multiple parts in one fzpz

KjellMorgenstern:
Note: Before Fritzing 0.9.5, these files would kind of load, but
show an error message and the parts would not be shown in "my parts",
but still be available via the search.


KjellMorgenstern:
Example
[b3f.fzpz.gz](https://github.com/fritzing/fritzing-app/files/4197783/b3f.fzpz.gz)
contains seven related parts that share the same pcb and schematic (but differ in description and breadboard)

KjellMorgenstern:
Rebasing and merging despite broken windows build. It is broken due to Qt policy changes, needs a separate fix.

PR #3635, opened 2020-03-01
Add spanish translations

albfan:
Trying to create a custom part I get blocked by: "There are no layers defined in %1"

https://github.com/fritzing/fritzing-app/blob/develop/src/partseditor/pemainwindow.cpp#L1630

Seems that it needs some groups with specific names "copper0" or "copper1". Maybe that could be explicitly said on warning message. I can provide a different patch if that makes sense.

I improved a little bit the spanish translations too.

To continue with improvements would it be possible to order colors in wires alphabetically by translated text? Is hard to find colors not alphabetically sorted (I can provide a different patch if that makes sense too)

### DIFF
--- a/translations/fritzing_es.ts
+++ b/translations/fritzing_es.ts
@@ -106,7 +106,7 @@
     <message>
         <location filename="../src/help/aboutbox.cpp" line="145"/>
         <source>Chinese (Traditional): </source>
-        <translation type="unfinished"></translation>
+        <translation>Chino (tradicional):</translation>
     </message>
     <message>
         <location filename="../src/help/aboutbox.cpp" line="145"/>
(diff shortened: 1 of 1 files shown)


translations/fritzing_es.ts
```diff
@@ -2507,12 +2507,12 @@ Es también posible de escoger un conector como tierra llena semilla por correct
...
         <location filename="../src/mainwindow/mainwindow_export.cpp" line="1029"/>
         <source>Export the current sketch to Extended Gerber format (RS-274X) for professional PCB production</source>
-        <translation type="unfinished">Exportación el esquemático actual a Extendido Gerber formato (RS-274X) para profesional PCB producción</translation>
+        <translation type="unfinished">Exportar el boceto actual a formato Gerber extendido (RS-274X) para producción profesional PCB</translation>
     </message>
     <message>
         <location filename="../src/mainwindow/mainwindow_export.cpp" line="1033"/>
         <source>Export the current sketch to PDF for DIY PCB production (photoresist)</source>
-        <translation type="unfinished">Exportación el esquemático actual a PDF para PCB al ácido</translation>
+        <translation type="unfinished">cococococoasdfaExportación el esquemático actual a PDF para PCB al ácido</translation>
```
KjellMorgenstern:
```suggestion
        <translation type="unfinished">Exportación el esquemático actual a PDF para PCB al ácido</translation>
```

KjellMorgenstern:
Translation updated where included in PR #3651

albfan:
Ouch. thanks!

PR #3654, opened 2020-04-10
Fix libgit2 version check to allow building with versions >= 1.0

PsikoBlock:
Until now, the check only considers the minor version number, so it breaks with versions greater than 0.99.


KjellMorgenstern:
Thanks for the PR.
Do you already use Fritzing with libgit 1.0, or is the change just to meet the conditions of your build env?

Issue #3647, opened 2020-03-17
design rules check not seeing all of the copper for specific part


mMerlin:
## Current Behaviour

Fritzing design rules check is not detecting collision between traces and the elongated portion of the pads for the LED Matrix parts I added to core recently.  It does detect collisions with identical looking pads in the MAX7219 and sparkfun 556 time chips.  I am not seeing an obvious (functional) difference between the pcb image implementations.

The working chips:
* copper1 and copper0 groups separate (so duplicate connector«n»pad elements)
* path element that extends the copper pad immediately after each pad circle
* a portion of the path element uses an absolute move
* no transform attributes
* stroke, stroke-width, fill attributes are on each element
* the silkscreen is created from a series of line elements plus a single path
* the pads are (mostly) outside of the silk screen

LED Matrix with narrow pads:
* copper0 nested inside of copper0
* extending path elements in a group after connectorpad circles (but still inside copper1 and copper0 groups)
* only relative coordinates are used in the path (all of the copper paths are identical, except for the initial move to position them)
* outer copper1 group has a translate to align the first pad center to 0,0
  * group of extended copper paths has a translate to adjust the path origin to 0,0 (so the paths have the same m«ove» coordinates as the circles have x and y)
  * the second row of pads (and extended copper) have a translate to move over 600mils
* stroke, stroke-width, fill values are set by a style attribute on the containing group
* the silkscreen is created from a single rect element
* the pad are completely in side the silkscreen

In the screen shots, the top row of pads is for the LED Matrix.  The bottom row, left side is a MAX7219, the right is a 556 timer chip.  From the **red** areas, frc is only detecting the pad circle element as a collision candidate for the Matrix, not the oblong pad around it.

testing sketch file (renamed)
[design rules check research-01.fzz.zip](https://github.com/fritzing/fritzing-app/files/4341111/design.rules.check.research-01.fzz.zip)

0.9.4b collision screenshot
![copper-collision-detection-0 9 4b](https://user-images.githubusercontent.com/918823/76814783-5aec3500-67c1-11ea-9899-3dc7b9dcc236.png)

0.9.5 collision screenshot
![copper-collision-detection-0 9 5](https://user-images.githubusercontent.com/918823/76814797-60497f80-67c1-11ea-8549-8a5de612e8c0.png)

So this says there is some combination of:

* a bug in the design rules check code about what it sees as (top and bottom) copper
* a real problem with the LED matrix part that I can not see
[...]

PR #3663, opened 2020-04-18
Svg parse tests

mMerlin:
Unit tests for the low level svg path parsing Fritzing classes. The failures from test_pathparse.cpp are the important ones. That is the cause for #3647. The others are (currently) cosmetic, and avoided by preprocessing before parsing starts.

### DIFF
--- a/tests/auto/test_svg/test_pathlexer_protected.cpp
+++ b/tests/auto/test_svg/test_pathlexer_protected.cpp
@@ -1,0 +1,50 @@
+#include "svg/svgpathgrammar_p.h"
+
+/*
+Test how SVGPathLexer::lexer cleans up (pre-processes) various svg path element
+data strings
+
+Testing that needs access to protected members of the class under test. This
+is not the best way to do this, but it is simple, and is fine for the
+functionality cases. Can break if class member allocation order is important
+*/
+
+// Get access to protected members of class SVGPathLexer for testing
+#define protected public
+#include "svg/svgpathlexer.h"
+#undef protected
+
+#include <boost/test/unit_test.hpp>
+
+BOOST_AUTO_TEST_CASE( pathlexer_clean )
+{
+	const QStringList test_cases = {
+		"m0,0",
+		"m 0 ,  0 \t\n   ",
+		"m0,0x",
+		"m 0 , 0 x",
+		"m 0 0\nx\n",
+		"m0,0a 2.6,2.6 0 0 1 5.2,0v5.2a 2.6, 2.6 0 0 1-5.2,0 z "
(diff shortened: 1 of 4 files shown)


mMerlin:
Hmmm, this is now show the already merged commits for boostdetect. I thought I had the branch synchronized, but apparently not. Is this pr good as is, or do I need to do some cleanup (and what)?

KjellMorgenstern:
I think that might be because the branch was  synchronized with a merge. Can you rebase the branch on origin/develop, so we get a fast forward commit? Thank you a lot.

mMerlin:
I could not think of a way to get rebase to get things back where they belong.  I had messed up my commit working branch commit history with a merge prior to adding the new content. I add to get a clean branch, then reapply the changes.  I need to check my notes about the process for synchronizing with upstream.  I **think** the problem surfaced because I did the work for boostdetect in my local develop branch, instead of creating a working/feature branch. Then my sync (merge) with upstream/develop after the pr was approved messed up the local develop history chain. Should be cleaned up now, and I also know to be more careful about where I do the work, and how I synchronize with upstream. Note to self: pay attention if you see a merge commit message when expecting a fast forward.

KjellMorgenstern:
I build and ran the test, currently it fails with with 13 errors:
*** 13 failures are detected in the test module "SVG Tests"

I guess that is expected. Do you plan to fix those errors in later commits, or skip them in the long term and fix later, whenever possible? The idea of course is to run those tests from the CI.

mMerlin:
Yes the failures are expected, and the pending changes would fix them.  Well fix with an additional (intended) change to the testing code itself.  That was supposed to be ready to go, but as mentioned in another thread, fixing the parsing error uncovered another error. Which needs its own test case, and might cause more problems than leaving the parsing error in place. The parsing error causes some copper to be ignored in the drc, the other causes (in certain situation) drc failure to be reported when it shouldn't.  A false positive instead of a false negative.  Currently trying to figure out how pcb svg data is loaded into itembase, so I can track from itembase which svg element ends up where during the drc. I know the processing is wrong, and seems to be because some `notnet` copper is not being merged with `net` copper for the pads. But have not been able to trace through the recursive itembase structure to see where that happens. I need a way to locate the problem item, so that I can track the logic that assigns it to the group.

If ci is going to fail with unit test errors being reported, this can wait, and be included with the fixes. Or, since this **IS** the development environment, I can either add to this pr, or create a new one to fix the parsing error (and pass the tests), then continue to chase the other error.

KjellMorgenstern:
No, not yet an issue, just wanted to make sure that there is a perspective for those tests to become green.

PR #3664, opened 2020-04-27
fix svg path element parse and matching of extra graphics to pcb conn…

mMerlin:
fixes for #3647. Get this (the actual fixes) in place, then I'll push additional documentation and minor refactoring changes worked on while researching where and what was causing the problems. This also clears the failures introduced with the previous unit test commit.

### DIFF
--- a/src/svg/svgpathlexer.cpp
+++ b/src/svg/svgpathlexer.cpp
@@ -21,18 +21,17 @@ along with Fritzing.  If not, see <http://www.gnu.org/licenses/>.
 #include "svgpathlexer.h"
 #include "svgpathgrammar_p.h"
 #include "../utils/textutils.h"
-#include <qdebug.h>
 
-static const QRegExp findWhitespaceBefore(" ([AaCcMmVvTtQqSsLlVvHhZz,])");
+static const QRegExp findWhitespaceBefore(" ([AaCcMmVvTtQqSsLlVvHhZzx,])");
 static const QRegExp findWhitespaceAfter("([AaCcMmVvTtQqSsLlVvHhZz,]) ");
 static const QRegExp findWhitespaceAtEnd(" $");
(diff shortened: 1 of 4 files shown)


src/svg/svgpathlexer.cpp
```diff
@@ -21,18 +21,17 @@ along with Fritzing.  If not, see <http://www.gnu.org/licenses/>.
 #include "svgpathlexer.h"
 #include "svgpathgrammar_p.h"
 #include "../utils/textutils.h"
-#include <qdebug.h>
 
-static const QRegExp findWhitespaceBefore(" ([AaCcMmVvTtQqSsLlVvHhZz,])");
+static const QRegExp findWhitespaceBefore(" ([AaCcMmVvTtQqSsLlVvHhZzx,])");
```
KjellMorgenstern, line 25: What is the "x" for? I couldn't find this in the TinySVG 1.1 standard, https://www.w3.org/TR/SVGTiny12/paths.html
but it could be part of a whitespace
```
wsp ::=
    (#x20 | #x9 | #xD | #xA)
```

mMerlin: 'x' is not part of svg at all.  It was added to the parser as a dummy entry to force an eof condition.  It should not be needed at all, but (testing for) the refactoring to remove it is a bigger job than I want to get into yet. Safer to leave it in at this time. See references to SVGPathLexer::FakeClosePathChar. https://github.com/fritzing/fritzing-app/blob/308f8d531d251c0b4474cdf9b76f1dffb89dad49/src/svg/svgfilesplitter.cpp#L822-L846
In this context, it gets treated almost the same as a z.  Almost, because an x always ends the command sequence. a z can have more path content after it. I do not like that code, but not sure enough about side-effects to make major changes to it.

PR #3666, opened 2020-05-01
doxygen, comments, code style for collectEqualPotential

mMerlin:
Cleaned up notes for one function inspected while researching cause for drc failures.

This is an initial sample to discuss, before I start doing pull requests for others I had my fingers in while researching the bugs. This one does not overlap the actual changes, so can be done before or after the pending bug fix pull request. Added the basic header that doxygen uses, a few comments to help keep track of context reading the code, and adjustments to match the published coding style.

### DIFF
--- a/src/connectors/connectoritem.cpp
+++ b/src/connectors/connectoritem.cpp
@@ -1304,53 +1304,67 @@ bool ConnectorItem::isConnectedToPart() {
 	return false;
 }
 
-void ConnectorItem::collectEqualPotential(QList<ConnectorItem *> & connectorItems, bool crossLayers, ViewGeometry::WireFlags skipFlags) {
-	// collects all the connectors at the same potential
-	// allows direct connections or wired connections
-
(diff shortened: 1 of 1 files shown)


KjellMorgenstern (approved):

KjellMorgenstern:
Improvements in the documentation are welcome. But notice how hard it is. If in doubt about the original developers intentions, it is better to leave the code without comment than adding a potentially inaccurate comment. That said, I think my concerns don't apply to this particular excellent PR.

mMerlin:
Thanks. This was not actually that difficult to **do**. The hard part was keeping from changing too much. To stay within the scope of the intended pr. To add documentation with VERY minor cleanup within the published coding style information.

For the "in doubt" cases, I might add a comment that is a question about the code intent. To provide a tag to go back to when more is learned. By me, or someone else. With the same guidelines for being useful. A comment that says "what does this do?" is not useful. One that lists what is known about the context, then asks a question about the expected path is more so.

Issue #3713, opened 2020-08-28
2.2mF capacitor


thomasw-mitutoyo-ctl:
## Problem
We have 1mF and 4.7mF capacitors, but not the 2.2mF capacitor.
IMHO 2200µF capacitors are also common

## Proposed Solution
Add another value to the dropdown list

KjellMorgenstern:
PR #3717 could be a fix, however it adds more values than just 2.2mF ( = 2200µF ) .

Issue #1455, opened 2014-08-25
missing resistor values, only E6 series available


davidperrenoud:
_From [stl...@poczta.fm](https://code.google.com/u/103015841192184689122/) on May 24, 2011 17:14:19_

What steps will reproduce the problem? 1. Take a resistor
2. Try to find 2.7kΩ

What is the expected output?
No such value on the list.

What do you see instead?
2.2kΩ and 3.3kΩ are available What version of Fritzing are you using? On what operating system? 0.5.2 (b4756) for Linux Please provide any additional information below. It looks like only E6[1] series is available (20&#37; tolerance, no fourth strip). I think it isn't uncommon to use E12 (10&#37;, silver strip). What's more misleading is that there is a golden strip on resistors that suggests E24 (5&#37;) series is available.

[1] http://en.wikipedia.org/wiki/Preferred_number

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=1455_

failiz:
E12 will create the too many values in the combo box. The proposed solution of two combo boxes (one for the number and one for the multiplier) will make the selection more complex for beginners and more time consuming. I would keep the E6 values in the combo box. The E12 values can be typed in and they are added to the combo box automatically.

KjellMorgenstern:
I also agree with typing in. There could be improvements for that, like remembering the last use values, or a 'custom' option in the combo box, that points user to the possibility to type in values.

PR #3717, opened 2020-09-14
added more capacitor values, fixes 3713.

failiz:
Hi,
I am not sure if I should add some of these values to the properties capacitance5mm, capacitance7.5mm, etc. When are they used? 
Cheers,
Andres

### DIFF
--- a/resources/properties.xml
+++ b/resources/properties.xml
@@ -78,10 +78,20 @@
 	<suffix suffix="CapacitorModuleID" />
 	<menuItem value="0.000000000001" />
 	<menuItem value="0.000000000010" />
+        <menuItem value="0.000000000022" />
+        <menuItem value="0.000000000047" />
 	<menuItem value="0.000000000100" />
+        <menuItem value="0.000000000220" />
+        <menuItem value="0.000000000470" />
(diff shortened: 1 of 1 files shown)


KjellMorgenstern:
You can see the usage of the capacitance5mm ... tables when looking for "foil capacitor" in the parts search.

KjellMorgenstern:
Do you think the values added are very common?
The original issue #3713 only asked for a 2.2mF capacitor.
 It is possible to enter a custom number to the capacitance field in case the value is not in the table. So, the table just is for convenience, and maybe to help beginners finding their capacitor. If the list includes rare settings, it might be less helpful.

failiz:
I do not have strong feelings about this. 
My intention was to provide all the [E3 series values](https://en.wikipedia.org/wiki/E_series_of_preferred_numbers#E6). I checked in RS (Amidata) and there are plenty of these values. However, I do not know if there are very used. 
This issue is related to #1455. The resistor implement the E6 series. 
Let me know if you prefer to only add that specific one.

KjellMorgenstern:
E3 sounds good. And there are the issues with manually entering a value (it works but somehow you need multiple attempts).

PR #3719, opened 2020-09-17
fixes #3616

### LINKED ISSUE #3616: Two parts connected via breadboard don't route
## Current Behaviour

Connecting two parts to each other via the breadboard does not route correctly.

[Two-parts-via-breadboard.fzz.zip](https://github.com/fritzing/fritzing-app/files/4172288/Two-parts-via-breadboard.fzz.zip)

(remove the trailing .zip to recover the .fzz file)

The two Neopixel parts are connected to each other in schematic but show as unrouted in breadboard:

Capture

Clicking on the unrouted message gets no unrouted nets:

Capture1

**Build:**
Version 0.9.4 
(CD-498-0-a1ffcea 2019-12-01) 64 [Qt 5.12.3]

**Operating System:**
<Windows 7>

**Steps to reproduce:**
- Connect pins on any two parts via a breadboard strip

## Expected Behaviour

Should be a routing complete message instead.

Capture2

failiz:
This commit fixes #3616 
I hope that this does not break anything or causes performance issues (now all the connections of the breadboard are analysed).

### DIFF
--- a/src/connectors/connectoritem.cpp
+++ b/src/connectors/connectoritem.cpp
@@ -1417,9 +1417,11 @@ void ConnectorItem::collectParts(QList<ConnectorItem *> & connectorItems, QList<
 		case ModelPart::Board:
 		case ModelPart::ResizableBoard:
 		case ModelPart::Via:
+		case ModelPart::Breadboard:
 			collectPart(connectorItem, partsConnectors, viewLayerPlacement);
 			break;
 		default:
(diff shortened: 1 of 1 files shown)


KjellMorgenstern:
Sorry for this delay. "hope it doesn't break" is not enough to include it. Unitl now I didn't find time for a deeper look at this, but form a first glance that looks like the proper solution.

KjellMorgenstern (approved):
Tested with more complex circuits, and with "Show unrouted" as well as clicking on routing status in th status bar.

Issue #3073, opened 2015-04-14
Export to png does not update until restart fritzing and reload file.


SzymonZy:
I jut notice that if you export schematic to PNG, change something, export again ... output fill will not include changes made since last export. 

You have to turn off fritzing, reload file and export again. That will work. I'm using 0.9.2.

thomasw-mitutoyo-ctl:
Still the same behavior in 0.9.4,

failiz:
Could you provide with detailed steps to reproduce it?
I can only reproduce it (0.9.4 in windows) when changing the name of a resistor. Steps to reproduce it:
1. Add a resistor in the schematic view of a new file 
2. Export to png
3. Double click in R1, set label to R2 and click ok
4. Export to png and the image still says R1

The name of the resistor updates only after: adding a second resistor, draw a wire between the resistors and move the first resistor.

thomasw-mitutoyo-ctl:
Steps for me:

0. Create a new untitled sketch
1. Add a resistor and have it selected (see the selection rectangle)
2. Export as PNG. Exported PNG will have the resistor selected
3. Select the breadboard (selection rectangle changes)
4. Export as PNG. Confirm overwriting the existing image or choose a different file name

Result: the PNG file still has the resistor selected

![grafik](https://user-images.githubusercontent.com/25632350/95726983-e48db380-0c79-11eb-802c-07b628aceaaf.png)

thomasw-mitutoyo-ctl:
Related code: 

mainwindow_export.cpp, line 970: creation of the button, associate with doExport()
mainwindow_export.cpp, line 481: doExport() method
mainwindow_export.cpp, line 548: calls exportAux()
mainwindow_export.cpp, line 556: exportAux() method

In sketchwidget.cpp, I find a lot of  ```// TODO: replace scene()->items()```. Why should it be replaced? Does it affect export as well? Should it be replaced there, too?

KjellMorgenstern:
Fixed by PR #3723 . Note that the export method still can not deal with the selection. It might still use a dangling pointer or at least a copy of the selection that should have been cleaned. The selection just doesn't reach the faulty part of the code anymore because the fix temporarily deselects.

failiz:
already released, close

PR #3723, opened 2020-10-09
Deselect all items before exporting an image to PNG or JPG. Fixes  #3712

### LINKED ISSUE #3712: Export PNG without selection rectangles
## Current Behaviour
Exporting the schematics includes the selection rectangles.

selection

**Build:**
Version 0.9.4 
(CD-498-0-a1ffcea 2019-12-01) 64 [Qt 5.12.3]

**Operating System:**
Windows 10 x64

**Steps to reproduce:**
1. Draw a schematic
2. Make sure some components are selected
3. Export schematics as PNG

## Expected Behaviour
The schematics is exported without considering selections, like so:
Kondensator Lade- und Entladekurve_Schaltplan

failiz:
The items are deselected, but they can be still be seen as selected in the app. After the screen is refreshed (f. ex. by zooming in the app or moving a component), they are shown correctly.

### DIFF
--- a/src/mainwindow/mainwindow_export.cpp
+++ b/src/mainwindow/mainwindow_export.cpp
@@ -557,6 +557,9 @@ void MainWindow::exportAux(QString fileName, QImage::Format format, int quality,
 {
 	if (m_currentGraphicsView == NULL) return;
 
+	//Deselect all items before creating the image
+	m_currentGraphicsView->selectDeselectAllCommand(false);
+
 	double resMultiplier = 3;
 
 	QRectF itemsBoundingRect;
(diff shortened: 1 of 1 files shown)


KjellMorgenstern:
Deselecting all items might be an unwanted effect. Do you see an easy way to preserve the users selection over the export? Or better, just not render that rectangles during the export?

failiz:
Now we deselect the items that are selected and after the export we select them again. Interestingly, this also fixes #3073 (the label of the resistor is exported well)

KjellMorgenstern:
I confirm #3712 and #3073 are fixed, looks good to me

PR #3727, opened 2020-10-19
The routing status algorithm needs to handle the part to part connect…

### LINKED ISSUE #2685: button example is not fully routed in bb view
_From [irasc...@gmail.com](https://code.google.com/u/104729248032245122687/) on July 26, 2013 08:10:20_

What steps will reproduce the problem? 1. 2. 3. Attach your sketch file and/or custom part files to the bug report. If the sketch uses custom parts, save the sketch as shareable (under the file menu). What is the expected output? What do you see instead? What version of Fritzing are you using? On what operating system? Please use labels and text to provide additional information.

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=2687_

failiz:
…ions (LED in an Arduino for example). Fixes  #2685

There are two things that I do not understand. First, why do we have two methods to calculate the routing status (updateRoutingStatus and showUnrouted)? Both of them do a similar task but how they do it is very different.
The second thing is why do we call three times updateRoutingStatus for each modification, one for each view (bb, schematics and pcb)? Would not be possible to calculate just the view where you are working and update the routing status when you switch to a different view?

### DIFF
--- a/src/utils/graphutils.cpp
+++ b/src/utils/graphutils.cpp
@@ -515,6 +515,16 @@ bool GraphUtils::scoreOneNet(QList<ConnectorItem *> & partConnectorItems, ViewGe
 			switch (toConnectorItem->attachedToItemType()) {
 			case ModelPart::Wire:
 				break;
+			case ModelPart::Part:
+				//Only the breadboard view allows part to part connections
+				if (toConnectorItem->attachedTo()->isEverVisible() && (myTrace & ViewGeometry::NormalFlag)) {
+					int j = partConnectorItems.indexOf(toConnectorItem);
+					if (j >= 0) {
+						add_edge_d(i, j, G);
+						add_edge_d(j, i, G);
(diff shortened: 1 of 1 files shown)


KjellMorgenstern:
Confirmed that this fixes the routing status for the digital input button example.

KjellMorgenstern:
Trying to answer the questions. The `showUnrouted` method is the slot to handle the signal from the menu, and it uses the 'collectAllNets' method. The status bar however uses the `updateRoutingStatuts` method.

Both methods have different output. While `connectAllNets` is usesd in various places, and outputs a list of subnets, the `updateRoutingStatus` discards that information. However, `updateRoutingStatus` has other side effects beyond counting the unrouted connections. I think it also removes orphaned ratsnest lines from the scene ( `ratsToDelete` list )

KjellMorgenstern (approved):

PR #3749, opened 2020-11-13
improvements to introduce values in the properties of the objects. 

### LINKED ISSUE #1262: hard to change capacitance values (probably incorrectly bailing on intermediate states)
_From [irasc...@gmail.com](https://code.google.com/u/104729248032245122687/) on October 07, 2010 10:50:35_

What steps will reproduce the problem? 1. 2. 3. What is the expected output? What do you see instead? Please use labels and text to provide additional information.

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=1262_

failiz:
The rules have been relaxed to allow intermediate values. The feasibility of the value is displayed in the background of the cell (red: unfeasible, green: feasible). Added a fixup function in case that the user presses enter or the combobox looses focus. The fixup function modifies the value to an acceptable value. The tip is more informative now, shows the range of feasible values for the property. Fixes #1262 
I am not sure if the fixup function should be change to set the min or max value if the value is out of range.

### DIFF
--- a/resources/properties.xml
+++ b/resources/properties.xml
@@ -225,23 +225,23 @@
 	<menuItem value="12.1" />
 </property>
 
-<property  name="maximum resistance" symbol="&#937;" minValue="" maxValue="" defaultValue="" numeric="yes" editable="yes">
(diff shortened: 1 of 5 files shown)


failiz:
This code was also in the simulator branch, which was merged to master in 0.9.10. Closing it.

PR #3758, opened 2020-11-22
More verbose error output when there is a problem with the parts db

KjellMorgenstern:


src/fapplication.cpp
```diff
@@ -309,13 +309,13 @@ void RegenerateDatabaseThread::run() {
...
+		m_error = tr("Unable to open temporary file") + " (" + fileName + ")";
 		return;
 	}
 
 	bool ok = ((FApplication *) qApp)->loadReferenceModel(fileName, true, m_referenceModel);
 	if (!ok) {
-		m_error = tr("Database failure");
+		m_error = tr("Database failure") + "\n" + m_referenceModel->error();
 		return;
 	}
```
KjellMorgenstern, line 321: Todo: can m_referenceModel be null

PR #3770, opened 2020-12-12
Mingw build support

jeremyd2019:
This pull request gets Fritzing building using mingw-w64, specifically tested under [MSYS2](http://www.msys2.org).

The first commit fixes gcc errors due to casting pointers to `long` which is not large enough to hold a pointer on 64-bit windows.  The second adapts the qmake files to basically treat `mingw` more like `unix` than `win32` (because the msys2 build environment is unix-like), and the third adjusts the search for the `applicationDirPath` to also act like `unix`, plus adds (bindir)/../share/fritzing as a candidate, as the msys2 mingw 'prefix' is relocatable in terms of Windows paths.

These patches were the result of getting Fritzing packaged in msys2/MINGW-packages#7376.

### DIFF
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -33,7 +33,7 @@ CONFIG += c++14
 # TODO: Verify flags for clang and msvc builds
 QMAKE_CXXFLAGS += -O3 -fno-omit-frame-pointer
 
-unix:!macx {
+if(unix:!macx)|mingw {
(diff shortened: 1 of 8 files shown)


Kreijstal:
thank you so much!!!!

Kreijstal:
@jeremyd2019 can you do it for latest fritzing version?

PR #3822, opened 2021-05-05
Clarify Dynamic libgit2 Instructions

### LINKED ISSUE #3821: Clarify Directions to Dynamically Include libgit2
## Problem
The current [comment in phoenix.pro](https://github.com/fritzing/fritzing-app/blob/a3d8cab74c48a67795d71c199c378317fe3aa633/phoenix.pro#L177) says:
```
# Disable this if you have (and want) libgit2 dynamically
LIBGIT_STATIC = true
```
It's ambiguous and some folks are going to comment this out instead of setting it to false, which leads to the qmake error:
```
libgit2detect.pri:22: if(condition) requires one argument.
```

## Proposed Solution
To save other folks some time, let's just rephrase it to: 
```
# Set this to false if you have (and want) libgit2 dynamically
LIBGIT_STATIC = true
```

merichar:
Replace the "Disable this" with "Set this to false" to prevent
commenting out the line and causing a qmake error.

Resolves #3821


janclod:
@merichar good idea :) I did incorporate your suggestion in my downstream [work](https://github.com/fritzing/fritzing-app/pull/3864), let me know if you have any remark.

PR #3831, opened 2021-05-21
Install app icon in XDG hicolor icon theme

pinotree:
Install the icon of the application in the hicolor XDG icon theme; this way it can be properly loaded by XDG menus in the currently set XDG icon theme, without looking in the legacy pixmaps location.

### DIFF
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -126,8 +126,8 @@ unix {
     manpage.path = $$DATADIR/man/man1
     manpage.files += Fritzing.1
 
-    icon.path = $$DATADIR/pixmaps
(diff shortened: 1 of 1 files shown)


pinotree:
@KjellMorgenstern PTAL :)

KjellMorgenstern:
Reading into this here https://doc.qt.io/qt-6/appicon.html first gave me the impression that the pixmap directory is used mostly for gnome, while the icons directory is used on kde.

https://specifications.freedesktop.org/icon-theme-spec/0.13/#install_icons  suggests to use further subdirectories for the theme, like in this PR.

Issue #3084, opened 2015-05-13
Application freezes and crashes


spelare69:
When starting a new design and the steps below are followed the application freezes and stops responding. 
1. New design
2. breadbord tab is selected
3. delete default breadboard
4. Add single-sided stripboard by draging from core parts
5. Set size of stripeboard to columns = 39 and rows = 30
6. search for Component atmega328 and then drag Component to stripboard
7. Applicationa freezes and stops responding as soon as the Component touches the stripboard
   As long as Component is not draged on top of the stripboard there is no problem.

This freezing is the same every time and I have tried 7-8 times.
No problem with smaller stripboard (default setting 30 columns x 20 rows) but when changing smaller stripboard design size to size 39 columns x 30 rows the application frezzes and stops responding.

Computer system:
windows 7 32-bit with service pack 1
i5 - 4670 3.40 GHz, 4GB of RAM

BR
Spelare69

aknoerig:
I'm afraid that's due to the relatively naive implementation we made. It's probably not freezing, but just taking a lot of time to compute the stripboard behaviour. More optimization work is needed here.

Gemba:
This behavior is caused by #3090 (see there for a workaround). It is unrelated to the selected component. Verified with 0.9.2b on Linux.

thkfighter:
Fritzing is freezing. It is fritzing 0.9.3b 64bit on Windows 10 Home Edition (upgraded to the newest on Oct 26, 2018). Each time I start fritzing, the application promts "not responding" or just freezes without any prompt.
When I run fritzing as the administrator, it does not freeze at the start but do after I do some operations like the ones mentioned by @spelare69 .

PC    Lenovo 20KFA01PCD laptop
OS    Windows 10 64bit home	
CPU  Intel Core i5-8250U @ 1.60GHz
Motherboard    Lenovo 20KFA01PCD ( Intel Xeon E3 - 1200 v6/7th Gen Intel Core/DRAM Registers - 5914 - 7th Generation Intel...)
RAM    8GB
HD    Sandisk SD9TN8W256G1001 ( 256 GB / SSD )
GPU  UHD Graphics 620 ( 128 MB )
Network card    Intel Ethernet Connection  I219-V

KjellMorgenstern:
Gemba is right: Because of #3090 all connectors of the board are connected to each other. This is not a typical case for a stripboard, and should be avoided. Furthermore, when Fritzing tries to build a list of connected items, it uses said naive implementation.

Issue #2998, opened 2014-09-23
Resize stripboard is adding strips the wrong way


aknoerig:
_From [Martin.B...@gmail.com](https://code.google.com/u/100779990985978385285/) on September 10, 2014 16:45:22_

What steps will reproduce the problem? 1. Drag a stripboard to Breadboard view
2. Resize it to e.g. 30 x 30 (becomes very slow)
3. Original stripboard looks fine, the added holes are not stripboard Attach your sketch file and/or custom part files to the bug report. What is the expected output? What do you see instead? Expected output is a larger stripboard. Looks like added holes are added the wrong way What version of Fritzing are you using? On what operating system? Version 0.9.0 (bda4e 2014-06-11) 64 [Qt 4.8.5] on Linux (PCLinuxOS 64 bit) Please provide any additional information below.

**Attachment:** [Reed.fzz](http://code.google.com/p/fritzing/issues/detail?id=2998)

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=2998_

GregWoods:
Is Fritzing dead? This bug is still present, 3 years later

SilverNodashi:
Yes it's dead

failiz:
@SilverNodashi  @GregWoods  dead revive sometimes :-)
remove imported label, add stripboard label

hxlnt:
+1, please fix. (Are repo owners hiding user comments on this bug? If so--why?)

failiz:
@hxlnt , the comments were out of topic, but owners are not hiding anything, you can press the button "Show comment" to see them.
In any case, I just pushed a fix for the issue. Lets hope that it is accepted and does not create new errors. Until is merged and there is a new release, you can see #3090 for a handy workaround.

georgyfarniev:
The issue is still here, in 2021. Feels like fritzing is abandoned.

failiz:
@georgyfarniev , Fritzing development was frozen for a few years and restarted two years ago. 

This bug has been fixed, and it will be released in 0.9.9. If you are able to compile Fritzing, you can test the fix in my pull request. If not, check  #3090 for a handy workaround.

PR #3856, opened 2021-07-11
resize a stripboard following the current layout,

### LINKED ISSUE #3090: Stripboard creation
when you resize a piece of stripboard the additional area has both horizontal and vertical strips.  It sorts itself out when you select a different orientation and then re select the desired one.

failiz:
fixes #3090 #2998 and #3084. The slow performance is still present but the extended area is not short-circuited.


KjellMorgenstern:
Closing, was merged with 24b6f99f6

Issue #3646, opened 2020-03-16
building libgit2 for Fritzing limitations


mMerlin:
## Current Behaviour

While setting up the Qt Creator environment to build/run/debug Fritzing, I was reviewing the build information for libgit2.  I noticed that it was using the bundled code for the http parser, even though http-parse was one of the dependencies included for libgit2-devel.  So I tried installing http-parser-devel.  That got the libgit2 build to show that it was no longer using the bundled code, **BUT**, when later attempting to build the phoenix project, I got a bunch or error messages just before the build failed.

```log
/usr/bin/ld: /home/phil/development/fritzing-app/../libgit2/build/libgit2.a(http.c.o): in function `clear_parser_state':
/home/phil/development/libgit2/src/transports/http.c:607: undefined reference to `http_parser_init'
/usr/bin/ld: /home/phil/development/fritzing-app/../libgit2/build/libgit2.a(http.c.o): in function `proxy_connect':
/home/phil/development/libgit2/src/transports/http.c:882: undefined reference to `http_parser_execute'
/usr/bin/ld: /home/phil/development/libgit2/src/transports/http.c:904: undefined reference to `http_errno_description'
/usr/bin/ld: /home/phil/development/fritzing-app/../libgit2/build/libgit2.a(http.c.o): in function `http_connect':
/home/phil/development/libgit2/src/transports/http.c:937: undefined reference to `http_should_keep_alive'
/usr/bin/ld: /home/phil/development/fritzing-app/../libgit2/build/libgit2.a(http.c.o): in function `http_stream_read':
/home/phil/development/libgit2/src/transports/http.c:1106: undefined reference to `http_parser_execute'
/usr/bin/ld: /home/phil/development/libgit2/src/transports/http.c:1132: undefined reference to `http_errno_description'
/usr/bin/ld: /home/phil/development/fritzing-app/../libgit2/build/libgit2.a(netops.c.o): in function `gitno_extract_url_parts':
/home/phil/development/libgit2/src/netops.c:236: undefined reference to `http_parser_parse_url'
collect2: error: ld returned 1 exit status
```

[...]

PR #3864, opened 2021-08-12
Fix issue #3646

janclod:
This should help future developers building on Linux using static libgit2 :crossed_fingers:


janclod:
If this pull request is merged, issue #3646 and #3821 may be closed :)

PR #3871, opened 2021-08-27
Add support for QuaZip 1.x

WOnder93:
QuaZip 1.x uses different include paths and ships with pkgconfig
metadata, so we can use PKGCONFIG instead of explicit library name. Use
the presence of QUAZIP_1X in defines to switch between QuaZip 1.x and
the legacy 0.x versions.


KjellMorgenstern:
Outdated, we always us 1.x now

Issue #684, opened 2014-08-25
rotating a part doesn't rotate its label


davidperrenoud:
_From [irasc...@gmail.com](https://code.google.com/u/104729248032245122687/) on June 01, 2009 10:58:20_

What steps will reproduce the problem? 1. 2. 3. What is the expected output? What do you see instead? Please use labels and text to provide additional information.

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=684_

failiz:
The labels still not rotate in the BB view when moving a part in 0.9.4. I think the labels should rotate when moving the part. However, the label should be always readable (the text should be horizontal and be able to read it from left to right or vertical and from bottom to top). Kicad labels are always horizontal, when you rotate a part. I think that having this as an option in the preferences menu is a lot of overhead for a minimal improvement. I would try to implement a good behaviour that works for most of us and not try to make everybody happy as the labels can still be changed manually. 
As a side note, I think that implementing hotkeys for rotating (parts or labels) could save a lot of time. Implementing the same ones as Kicad will make them easy to use for the users.

failiz:
My PR #3874 fixes this issue

KjellMorgenstern:
Now, with #3874, it will be so easy to rotate the label, that we don't need to rotate it along and keep Fritzings behavior, that the orientation stays readable by default?

failiz:
I am not following you...
#3874 has a couple of fixes. Add shortcuts to rotate and flip a part and keeps the labels readable. 
Regarding the shortcuts, they rotate a part. It is not possible to rotate just the label. That needs to be done with the right click contextual menu. 
One option would be not to rotate the label when rotating a part, but this would break the current behavior (except in bb view, but the labels are hidden by default in bb view). In Fritzing, when rotating a part, it rotates the label.

KjellMorgenstern:
So, you did not change the default rotation behavior in BB nor in the other views?

failiz:
I changed the behavior of the BB view to have the same behavior as the sch and PCB. I thought that it was odd to have different behaviors. And I took the sch and PCB one as the default as it is where you see the labels by default.

failiz:
This is the current behaviour in Firtzing 0.9.9 when rotating a part.

Schematics:
[...]

PR #3874, opened 2021-08-28
Added shorcuts for rotting, flipping and find part. Fixes #3582

### LINKED ISSUE #3582: Support CTRL-F for searching a component
## Problem
Many times I am searching for a component I use CTRL-F but the shortcut is not supported in windows.

## Proposed Solution
Add shortcut for searching a component

failiz:
When I made the examples for the simulator, making the circuits took too much time. These shortcuts will help to make Fritzing more user friendly. I used R and F (without control) as it is the same as in Kicad.

### DIFF
--- a/src/mainwindow/mainwindow_menu.cpp
+++ b/src/mainwindow/mainwindow_menu.cpp
@@ -951,6 +951,7 @@ void MainWindow::createPartMenuActions() {
 	connect(m_rotate45cwAct, SIGNAL(triggered()), this, SLOT(rotate45cw()));
 
 	m_rotate90cwAct = new QAction(tr("Rotate 90° Clockwise"), this);
+	m_rotate90cwAct->setShortcut(tr("R"));
 	m_rotate90cwAct->setStatusTip(tr("Rotate the selected parts by 90 degrees clockwise"));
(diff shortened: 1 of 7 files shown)


KjellMorgenstern:
I'd like to review the keys. Using the same keys as as KiCad could be a point, but my intuiton is  "[" and "]" for rotating and "|" "_" for flipping. One reason for it: I think man graphic tools use [ and ] (what does Inkscape use?)

failiz:
No Ida about Inkscape or man graphic tools, but it could be nice to check. By the way, I found after submitting this PR that there is already a shortcut for rotating in Fritzing (rotating the selection). It is CTRL+R and other variations also exist. As this action is not in the menues, it is not displayed in the menues and it is difficult to find (I found it by searching in the code). Maybe it should be added to the menu, but anly rotate 90 degrees. I do not think that people use the 45 degrees option so often.

KjellMorgenstern:
We added the search key, and 90 degree rotation with "[" and "]" . The label alignment in breadboard view was not changed, I think in most cases it is preferable to have the lables aligned, and rotating a single label is a special case.

failiz:
I am ok with not changing the lables in the BBview, but I think the labels should be readable in the SCH view at all times. I added some comments in #684.

PR #3883, opened 2021-09-22
fix power symbols which were broken by commit  #3845, fixes #3882

### LINKED ISSUE #3882: core parts schematic view power nets all connect together.
## Current Behaviour

From

https://forum.fritzing.org/t/power-symbol-connection-error/13887

On Fritzing 0.9.3b and 0.9.6 this works as expected (only 0.9.3b shown here!):

capture

capture1

on 0.9.8 the 3V3 and 5V nets are connected together which is incorrect.

capture2

capture3

**Build:**

Version 0.9.8 
(bCD-92-0-b3d4fefa 2021-08-03) 64 [Qt 5.15.2]

**Operating System:**
Windows 10

**Steps to reproduce:**
- Drag a 4001 subpart IC in to the sketch
- in schematic drag a 5V power connector in to the sketch 
- then drag a 3.3V power connector in to the sketch (or dup and change the 5V one) 
- drag a 2 pin header in to be the power source and connect the 3.3V to one pin and 5V to the other
- dup the 5V power net and attach it to pin 14
- dup the 3.3V power net and attach it to pin 2
- in pcb both headers and pins 14 and 2 all connect together which is incorrect. 

## Expected Behaviour
The 5V power pin should connect to pin 14 and the 3.3V power connector should connect to pin 2.

failiz:


KjellMorgenstern (commented):
Ups, we created a similar fix at the same time.
https://github.com/fritzing/fritzing-app/pull/new/fix3882

I didn't personally check any of the two variants, I guess both work.
But from looking at the code, I think the m_label Variable should have been implemented in NetLabel, which inherits SymbolPaletteItem.
It seems wrong that specialized behavior is implemented in the parent class. Similar applies for m_voltageReference.

If refactoring the m_label into the "NetLabel" class is feasible, I'd prefer it fixed this way. 
However, a working solution has priority before cleaning up code (there would be a lot to clean up if we start this). If you are interested in the refactoring, I'd wait a little, otherwise we can also use this fix.

failiz:
Funny that we made two fixes in parallel...
I also thought about refactoring the code, it looks more complex that it should be. But I think that the refactoring will take some time: The labels are also used by the SymbolPaletteItem class (SymbolPaletteItem::setLabel) and there is a variable called m_isNetLabel in SymbolPaletteItem. 

I am busy with other projects and I do not have time for the refactoring. So, choose the fix that you think is the best. I did not implement your approach as I thought that it would be better to check for voltage references in the getlabel function, but I guess that the other one works too.

KjellMorgenstern:
Fixed in 0.9.9

Issue #3627, opened 2020-02-25
Raise and Lower breadboard part incorrect until reopen sketch


mMerlin:
## Current Behaviour

With this sketch, moving a part backward on the breadboard view, moved a different part (the one that should have been the back most) to the front instead.  At least that is what it looks like on the screen.  However, an (immediate) svg export showed the requested part move instead.  Save, close, reopen got the breadboard view synchronized with what the previous export showed.

Screen shot and svg (with dummy .zip suffix, just rename to remove) export immediately after opening the sketch.

![wirewrap-eye1-sequence00a-bb](https://user-images.githubusercontent.com/918823/75296370-34526400-57ea-11ea-85d7-37bf9876d828.png)
[wirewrap-eye1-sequence00a-bb.svg.zip](https://github.com/fritzing/fritzing-app/files/4252621/wirewrap-eye1-sequence00a-bb.svg.zip)

Screen shot and svg (as before) export immediately after send backward

![wirewrap-eye1-sequence00a2-bb](https://user-images.githubusercontent.com/918823/75296597-d8d4a600-57ea-11ea-83e7-6df02edd24a3.png)
[wirewrap-eye1-sequence00a2-bb.svg.zip](https://github.com/fritzing/fritzing-app/files/4252628/wirewrap-eye1-sequence00a2-bb.svg.zip)

Screen shot after save, close, reopen

![wirewrap-eye1-sequence00b-bb](https://user-images.githubusercontent.com/918823/75296681-1cc7ab00-57eb-11ea-9f2a-1df57ccc12f2.png)

I have had regular issues manipulating parts using Raise and Lower.  This was a case that I could reproduce, and the research done gathering the details for reporting discovered that it seems to be a display problem.  The SVG export showed what I would expect from the actual operation performed, so I expect the sketch content is correct as well.  But save, close, reopen after every Raise and Lower operation, to see if really got what was intended is not a nice work flow.

**Build:**

Version: 0.9.4b
Commit: c595162a1d06f01eab9b66092c7ca68518974b9887295cd59950c439bfba3a72
with flatpak

**Operating System:**

% cat /etc/os-release
NAME=Fedora
VERSION="31 (Workstation Edition)"
ID=fedora
VERSION_ID=31
VERSION_CODENAME=""
PLATFORM_ID="platform:f31"
PRETTY_NAME="Fedora 31 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
[...]

Issue #363, opened 2014-08-25
 Move the developer instructions on the mailing list to the Google Code project wiki 


davidperrenoud:
_From [leidel](https://code.google.com/u/leidel/) on December 11, 2008 05:38:59_

while the mailing list is good for discussion of developer related things, the actual howto should be 
moved to the google code wiki system.

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=363_

PR #3886, opened 2021-10-02
fixes the send to back/front and raise and lower commands (#363…

### LINKED ISSUE #3632: Move part to back does nothing, or does the wrong thing
## Current Behaviour

Using Raise and Lower, Move to Back in the breadboard view does nothing at all for one part (not the back one) in this sketch, and messes up the z ordering when used on another part.

Moving U1 to back did not change the z ordering at all.  The raw part geometry was extracted before and after to verify.

Moving R1 to back changed the z geometry value of all parts.  3 were changed to exactly 2.  2 more, include R1 were set to 2.00001, and the final part was set to z=2.00002  Even ignoring the duplicate values, none of those are correct based on the previous z values.  R1 should have ended up the lowest, and J1 should have remained the highest.  If the z value for R1 were the only thing changed, the ordering would have been correct.

Before attempting to move U1 to the back (extra leading white space removed)

```xml
grep -A12 -e "instance module.*path=\"/home" -e "logotext\.fzp" raise-lower-test01.fz | grep -e "geometry z" -e "title"
<title>U1</title>
  <geometry z="2.50007" x="0" y="16.7098">
<title>R2</title>
  <geometry z="2.50003" x="79.7783" y="31.021">
<title>Disp1</title>
  <geometry z="2.50002" x="18" y="0"/>
<title>R1</title>
  <geometry z="2.5001" x="-10.2316" y="31.0752">
<title>J1</title>
  <geometry z="2.50014" x="9.00314" y="0.001458"/>
<title>TXT1</title>
  <geometry z="2.50008" x="0.210197" y="0.274506"/>
```

After attempting to move U1 to the back

```xml
grep -A12 -e "instance module.*path=\"/home" -e "logotext\.fzp" raise-lower-test01.fz | grep -e "geometry z" -e "title"
<title>U1</title>
  <geometry z="2.50007" x="0" y="16.7098">
<title>R2</title>
  <geometry z="2.50003" x="79.7783" y="31.021">
<title>Disp1</title>
  <geometry z="2.50002" x="18" y="0"/>
<title>R1</title>
  <geometry z="2.5001" x="-10.2316" y="31.0752">
<title>J1</title>
  <geometry z="2.50014" x="9.00314" y="0.001458"/>
<title>TXT1</title>
  <geometry z="2.50008" x="0.210197" y="0.274506"/>
```

After moving R1 to the back

```xml
<title>U1</title>
  <geometry z="2" x="0" y="16.7098">
<title>R2</title>
  <geometry z="2.00001" x="79.7783" y="31.021">
<title>Disp1</title>
  <geometry z="2" x="18" y="0"/>
<title>R1</title>
  <geometry z="2.00001" x="-10.2316" y="31.0752">
<title>J1</title>
  <geometry z="2" x="9.00314" y="0.001458"/>
<title>TXT1</title>
  <geometry z="2.00002" x="0.210197" y="0.274506"/>
```

Those z values started as 2.5, 2 zeros, 2 digits.  They ended up as 2.0, 2 zeros, 2 digits, with the final part messed up.  That looks like poor management of the z values.  I have an early sketch snapshot that lead to this version.  There were only 3 parts in that with z values of 2, 2.00001, 2.00002.  It looks like that with the 2.5… values, something may have triggered a resequencing process, which messed up.

I have worked with other code that needed to be able to do key insertions.  Normally those use a larger increment when adding sequential entries, to allow room to insert between without a full resequencing.  The range seen here contains 50014 values.  That is a **LOT** of parts.  Incrementing by 500 during placement, then inserting in the middle of the gap would likely reduce the need to adjust multiple values at once.  Or for single step forward and back, swap the 2 values.  Doing that, and starting in the middle of the available range (2.25000 here) should make it all but impossible to need to cleanup the sequencing.  Swaping 2 adjacent values is always safe  So is taking any one entry, and setting to be one lower that the current minimum, or one higher than the current maximum.  It would take 25000 move to back operations to run out of that space.  Problems could be encountered sooner, if there were options to add the next part before or after a specific part.  Which would then go back to using a larger increment to leave room most of the time.

Not tested, but some symptoms indicate that cloning a part will put it in the z order adjacent to the original part.  Which would then need to have, or create, a gap.

**Build:**
app/org.fritzing.Fritzing/x86_64/stable
Version: 0.9.4b
Commit: c595162a1d06f01eab9b66092c7ca68518974b9887295cd59950c439bfba3a72

**Operating System:**

```sh
% cat /etc/os-release
NAME=Fedora
VERSION="31 (Workstation Edition)"
ID=fedora
VERSION_ID=31
VERSION_CODENAME=""
PLATFORM_ID="platform:f31"
PRETTY_NAME="Fedora 31 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:31"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f31/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=31
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=31
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

**Steps to reproduce:**

[raise-lower-test01.fzz.zip](https://github.com/fritzing/fritzing-app/files/4266344/raise-lower-test01.fzz.zip)

* unzip raise-lower-test01.fzz
* `grep -A12 -e "instance module.*path=\"/home" -e "logotext\.fzp" raise-lower-test01.fz | grep -e "geometry z" -e "title"`
* open raise-lower-test01.fz; right click U1; Raise and Lower; Move to back; Save
* `grep -A12 -e "instance module.*path=\"/home" -e "logotext\.fzp" raise-lower-test01.fz | grep -e "geometry z" -e "title"`
* right click R1; Raise and Lower; Move to back; Save
* `grep -A12 -e "instance module.*path=\"/home" -e "logotext\.fzp" raise-lower-test01.fz | grep -e "geometry z" -e "title"`

## Expected Behaviour

The z ordering should change when moving U1 to the back.

When moving R1 to back, either only the z value of R1 should change, or if multiple changed, the relative z order of the other parts should not change, and R1 should end up the lowest.

failiz:
(fixes #3632 and #3627)


failiz:
I'm not so sure about the second commit. It fixes the wrong z values but it has not been tested in other files or pcb views. Not sure if I'm missing something. 
In any case, the problem is why the z values get wrong values. This, this PR is just a workaround.

failiz:
Released in 0.9.10. Closing.

failiz:
Released in 0.9.10. Closing

PR #3896, opened 2021-10-22
fixes jump label behavior in right clicks (fixes #3895)

### LINKED ISSUE #3895: In schematic right clicking then moving a label causes a jump in position
## Current Behaviour

If you place a part in schematic, then right click on the label and move it slightly (1 grid line in this case) the position jumps a large distance.

capture

like this

capture1

if you left click on the label and move it it moves as expected only 1 grid line without jumping a large (and apparently random) direction,.

**Build:**

Version 0.9.9 
(bCD-348-0-f0af53a9 2021-09-22) 64 [Qt 5.15.2]

**Operating System:**
Windows 10

**Steps to reproduce:**
- See current behaviour

## Expected Behaviour
The label should move with the mouse as a left mouse click does not jump.

failiz:


KjellMorgenstern:
This was not reproducable, or fixed in a different way

failiz:
The issue is still there and it can be reproduced.

Issue #1326, opened 2014-08-25
check disconnect/reconnect behavior when changing voltage on a power symbol


davidperrenoud:
_From [tjtomk...@gmail.com](https://code.google.com/u/115783105138156191836/) on January 22, 2011 05:01:12_

What steps will reproduce the problem? 1. Create a small circuit using power.fpz
2. Modify power value to 3.3 or 12 volts
3. Try to save circuit

What is the expected output? 
Expected outcome is a saved circuit.

What do you see instead?
The application crashes.

What version of the product are you using?
Fritzing 0.4.4 (b4641)

On what operating system?
OSX 10.6.6 Please provide any additional information below.

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=1326_

failiz:
When changing the voltage in a power symbol, the wires which were connected to it are deleted. This still happens in 0.9.4. However, I think this is not the right behaviour. Users should known what they are doing and if you put the power pin, wire it and then change the voltage you have to wire it again.
The previous argument ("Because it could lead to a situation where some parts are connected to two voltages") does not seem coherent when you can connect two power symbols with different voltage to the same net and Fritzing does not complain at all.

failiz:
Remove imported label and add schematic view

failiz:
This is caused by function SchematicSketchWidget::setVoltage(double v, bool doEmit)
This is the code:
```

	QUndoCommand * parentCommand =  new QUndoCommand();
	parentCommand->setText(tr("Change voltage from %1 to %2").arg(sitem->voltage()).arg(v));

	new CleanUpWiresCommand(this, CleanUpWiresCommand::UndoOnly, parentCommand);
	new CleanUpRatsnestsCommand(this, CleanUpWiresCommand::UndoOnly, parentCommand);

	QList<Wire *> done;
	foreach (ConnectorItem * toConnectorItem, sitem->connector0()->connectedToItems()) {
		Wire * w = qobject_cast<Wire *>(toConnectorItem->attachedTo());
		if (w == NULL) continue;
		if (done.contains(w)) continue;

		QList<ConnectorItem *> ends;
		removeWire(w, ends, done, parentCommand);
	}

	new SetPropCommand(this, item->id(), "voltage", QString::number(sitem->voltage()), QString::number(v), true, parentCommand);

	new CleanUpRatsnestsCommand(this, CleanUpWiresCommand::RedoOnly, parentCommand);
[...]

PR #3897, opened 2021-10-23
fix disconnect behaviour (fixes fixes #2669 and #1326)

### LINKED ISSUE #2669: Changing Net Label name deletes wires
_From [bitsybof...@gmail.com](https://code.google.com/u/112850805511005419452/) on July 17, 2013 23:19:52_

What steps will reproduce the problem? 1. Drag label onto schematic
2. Connect it to a bend point on some wire between a couple of components
3. Change the label name
4. Observe that the complete wire you connected it to including through bendpoints is deleted and the label and components are not connected to anything anymore.

This occurs not just with new labels, renaming any label deletes the wire(s) connected to it right though bendpoints.

It is not be undo-able either, so rather annoying if you rename a label connected to a complicated wiring, and poof, all gone :-(

I don't think renaming the label should delete any wires, doesn't make sense to me to do that.

_Original issue: http://code.google.com/p/fritzing/issues/detail?id=2671_

failiz:
when changing voltage of power symbols and netlabel´s text, do not remove the wires attached to the symbol or the netlabel.

### DIFF
--- a/src/sketch/schematicsketchwidget.cpp
+++ b/src/sketch/schematicsketchwidget.cpp
@@ -260,24 +260,16 @@ void SchematicSketchWidget::setProp(ItemBase * itemBase, const QString & prop, c
 			QUndoCommand * parentCommand =  new QUndoCommand();
 			parentCommand->setText(tr("Change label from %1 to %2").arg(oldValue).arg(newValue));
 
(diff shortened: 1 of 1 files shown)


failiz:
Released in 0.9.10. Closing

Issue #3933, opened 2022-02-23
Connector issues in battery when shortcircuit


failiz:
## Current Behaviour

![image](https://user-images.githubusercontent.com/5708642/155310018-c6056131-18ea-48ec-bdc8-9b62ed04cd16.png)

Version 0.9.9, Windows 10

**Steps to reproduce:**
- Add a battery from core
- Connect the two terminals of the battery with a wire in schematic view
- Connectors are red, instead of green, but they are connected

## Expected Behaviour
Green connectors

failiz:
This happens for all the parts. It is the same issue as #3827

failiz:
Closing this issue in favor of #3827

Issue #3827, opened 2021-05-09
Connections between part pins in at least pcb don't show connected.


vanepp:
## Current Behaviour

In pcb view making a connection between pins on an IC (but not off chip) connect but don't show green for connected.

![bug](https://user-images.githubusercontent.com/16514277/117580306-37163d80-b0ac-11eb-85c2-01008538f033.jpg)

they correctly show up as rats nest lines in schematic:

![bug1](https://user-images.githubusercontent.com/16514277/117580354-6dec5380-b0ac-11eb-9548-890616d2b119.jpg)

and correctly show as connected if there is an off part connection:

![bug2](https://user-images.githubusercontent.com/16514277/117580413-aa1fb400-b0ac-11eb-8da6-4c763788012b.jpg)

**Build:**
Version 0.9.6 (bCD-175-0-8a1e0682 2021-02-21) 64 [Qt 5.12.10]

**Operating System:**

Windows 10

## Expected Behaviour

Image 1 should have green connections like image 3.

failiz:
Probably, the same bug as #3933

Issue #3580, opened 2019-12-16
Wires in schematic view are not drawn correctly if part is moved


NilsRo:
## Current Behaviour
In schematic view the wires are not drawn correctly if a parted is moved. It seems that the are not connected anymore but it is still fine. Only the wires drawn on the monitor are not correct anymore.

![2019-12-16_22-33-40](https://user-images.githubusercontent.com/51311453/70944836-2945bf80-2054-11ea-9257-1308135a0a48.png)

Video: https://readycloud.netgear.com/client/dllink.html#t=01h4idsbrhl2raau8xq76wxzkij/2019-12-16_22-30-06.mp4

[Sicherheitsschaltung Ladekoffer.zip](https://github.com/fritzing/fritzing-app/files/3970246/Sicherheitsschaltung.Ladekoffer.zip)

**Build:**
Version 0.9.4 
(CD-498-0-a1ffcea 2019-12-01) 64 [Qt 5.12.3]

**Operating System:**
Windows 10

**Steps to reproduce:**
- Move parts in schematics in the attached circuit.

## Expected Behaviour
Wires drawn correctly.

KjellMorgenstern:
A related problem exists in breadboard view.
Note: the displayed number of unrouted connections is probably not related to this issue (see #3581) .

NilsRo:
Hi, #3581 is not related to this. Today I ran into the problem that the drawing cannot be fixed and I cannot connect a wire anymore. 
![2019-12-17_12-54-49](https://user-images.githubusercontent.com/51311453/70993659-bcbad700-20cc-11ea-8ba5-c1623dad4823.png)
[Sicherheitsschaltung Ladekoffer.zip](https://github.com/fritzing/fritzing-app/files/3973139/Sicherheitsschaltung.Ladekoffer.zip)

KjellMorgenstern:
When I click the connector 1 of R32, it is highlighted in yellow, together with R31.1 , C12.2 and all the others of this subnet.  Further, if I move the bendpoint left of R32, then R32 might jump as well.

KjellMorgenstern:
Created a smaller version of the file that still has/triggers this error:
[simplified-3580.fz.zip](https://github.com/fritzing/fritzing-app/files/3974731/simplified-3580.fz.zip)

NilsRo:
[...]

PR #3942, opened 2022-03-24
Fixes dragging bendpoints of 4 or more wires (fixes #3006).

### LINKED ISSUE #3006: Connectors with with 4-Wires at node highlight but dont move together
Connectors with with 4-Wires at node highlight but dont move together. The disconnected portion "blips" between one side or the other when I move the node about.

screenshot - 09282014 - 07 41 57 pm

failiz:
This fixes issue #3006. When dragging a bendpoint with a junction of 4 or more wires, all wires are updated correctly. This fixes the schematic and PCB views. 
There is a similar problem  (issue #3580) when dragging a part (the wires connected to that part are not updated correctly). Probably, it is the same cause as in this bug, but in a different part of the code (dragging part code). There is a video of this issue also in #3006.

### DIFF
--- a/src/connectors/connectoritem.cpp
+++ b/src/connectors/connectoritem.cpp
@@ -535,6 +535,10 @@ void ConnectorItem::restoreColor(QList<ConnectorItem *> & visited)
 		}
 	}
 
+	//In PCB view, there are two connectors on top of each other.
+	//This line removes the other one which is in the opposite layer.
+	attachedTo.remove(this->getCrossLayerConnectorItem());
+
 	foreach (ConnectorItem * connectorItem, connectorItems) {
 		if (connectorItem->isEverVisible()) {
 			//QString how;
--- a/src/connectors/connectoritem.cpp
+++ b/src/connectors/connectoritem.cpp
@@ -530,14 +530,13 @@ void ConnectorItem::restoreColor(QList<ConnectorItem *> & visited)
 	foreach (ConnectorItem * connectorItem, connectorItems) {
 		if (connectorItem->isEverVisible()) {
 			if (connectorItem->attachedToItemType() != ModelPart::Wire) {
-				attachedTo.insert(connectorItem);
+				//For PCB pad stacks: Only add one connector to the attachedTo list (top or bottom copper)
+				if(!attachedTo.contains(connectorItem->getCrossLayerConnectorItem())) {
+					attachedTo.insert(connectorItem);
+				}
 			}
 		}
 	}
-
-	//In PCB view, there are two connectors on top of each other.
-	//This line removes the other one which is in the opposite layer.
-	attachedTo.remove(this->getCrossLayerConnectorItem());
 
 	foreach (ConnectorItem * connectorItem, connectorItems) {
 		if (connectorItem->isEverVisible()) {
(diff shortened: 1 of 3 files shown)


KjellMorgenstern:
The yellow highlight for same level subnet should still highlight all connectors of that breadboard strip. Is it affected by the patch?

failiz:
> The yellow highlight for same level subnet should still highlight all connectors of that breadboard strip. Is it affected by the patch?

No, that is handed by a different code and still works with my fix. My patch only modifies the colors of the connectors (green, red or no color) based on their connection status.

failiz:
Hi,
I just noticed that the dragging of the bendpoints of 4 wires has been fixed in 1.0.0. However, the second commit (580a8c4d02f88783f3a8636e22bf49d9d295c4d2) "Improvements to the connectors status color (red=unconnected, green=connected). Fixes https://github.com/fritzing/fritzing-app/issues/3933 and https://github.com/fritzing/fritzing-app/issues/3827 Now, when an item is connected to the breadboard, only the holes of the breadboard that are connected are in green; a wire connecting two holes of the breadboard has both ends in green." has not been merged. Any issues with the fix?

Edit: Notice that the commit fixes #3827

KjellMorgenstern:
It looks like with the patch applied, all connectors are marked "connected", even if none is:
image

failiz:
Yes, I tried it in the breadboard and schematic but I did not notice the error in the PCB view. The problem is that is finding two connectors (the top one and the bottom one). I have made a fix to consider only one of the connectors of the stack. Could you try it again to double-check it?

failiz:
image

I think it is better than before but there is something that does not work correctly in PCB. If you create traces that are not connected to anything and move the end of the unconnected traces, these traces are in green:
image
But when you move the component, then they all get a red color:
image

failiz:
Ok. I fixed the issue. Can you give it a try?
I think the behavior is better than the current one.

src/connectors/connectoritem.cpp
```diff
@@ -1365,7 +1386,7 @@ void ConnectorItem::collectEqualPotential(QList<ConnectorItem *> &connectorItems
 		// When the kept connector item is part of a bus, include all of the other
 		// connectors on the bus in the list being processed
 		Bus *bus = connectorItem->bus();
-		if (bus) {
+		if (bus && (followBuses||fromWire)) {
 			QList<ConnectorItem *> busConnectedItems;
```
KjellMorgenstern, line 1390: I think the "skipBuses" param was broken, or maybe never implemented.
For the IPC-D 356 export we also needed this and I modified this to
```
@@ -1381,7 +1384,7 @@ void ConnectorItem::collectEqualPotential(QList<ConnectorItem *> &connectorItems
                // When the kept connector item is part of a bus, include all of the other
                // connectors on the bus in the list being processed
                Bus *bus = connectorItem->bus();
-               if (bus) {
+               if (!skipBuses && bus) {
                        QList<ConnectorItem *> busConnectedItems;
```

The renaming is not an issue, but I am a bit paranoid about the "fromWire". The collectEqualPotential method is quite critical. Do you think by adding the skipBuses like in our current version, we are excluding to many items?

Is the patch about more than just highlighting the connections? Or do we need it to send proper netlists to ngspice?

failiz: I think skipBuses was never implemented and you implemented it recently. At least, I cannot find it in the source code. I guess skipBuses and followBuses are basically the same. 

The fromWire is because I want to avoid following buses of the parts (e.g., breadboard or Arduino pins), but I still want to follow the wires. And the wires are implemented as buses. Removing the fromWire from the code breaks the connector highlighting. 

The patch is to fix issue #3827 and is just about improving the highlighting of the connector status (not related to the SPICE netlist). That issue is caused because the current algorithm checks if a part is connected to another part (ItemBase). In my patch, I changed it to check if a connector is connected to other connectors (from the same part or from a different one). No following the buses makes that ony the hole of the breadboard that is connected is highlighted. 

In my code, the only point where buses are not followed is in the restoreColor function. The code for generating the spice netlists uses the default argument of followBuses (as it is necessary to follow the buses). Not sure how the exporting of the IPC-D 356 netlists works.

Issue #3914, opened 2021-12-15
Linkage error when building with vs2019


sevketcaba:
## Current Behavior
When building with vs2019, because of libgit2, linker throws linkage errors against
* Winhttp 
* Advapi32 
* Crypt32 
* Rpcrt4 
* Ole32

**Screenshot:**
![image](https://user-images.githubusercontent.com/18246855/146102766-6a587c6a-6b04-4469-8d81-b2ef2d92e5bc.png)

**Build:**
Commit : f0af53a9077f7cdecef31d231b85d8307de415d4
Qt 5.15.2
MSVC2019 (16.0)

**Operating System:**
<Windows 10>

**Steps to reproduce:**
- Build with above environment

## Expected Behavior
Build and link successfully

Issue #3595, opened 2020-01-20
Running qmake error while using libgit2-0.28.4.tar.gz 


arimogi:
## Error while building Fritzing
I try to build Fritzing from the source. I have installed `boost` and `libgit2` as said in https://github.com/fritzing/fritzing-app/wiki/1.-Building-Fritzing and located those directories as siblings fritzing-app directory. Also, I have followed https://github.com/fritzing/fritzing-app/wiki/1.3-Linux-notes.
While I ran `qmake` command, an error occurred as illustrated below:

![Screenshot from 2020-01-20 16-53-27](https://user-images.githubusercontent.com/1080574/72715348-46b5fe00-3bab-11ea-8e34-75e55b9c8146.png)

It says:
`Project MESSAGE: found libgit2 include path at /path/to/fritzing-app/../libgit2/include`
`Project ERROR: static libgit2 library not found in /path/to/fritzing-app/../libgit2/build`

**Operating System:**
<Linux Ubuntu 18.04>

**Steps:**
- Installed `boost`
- Installed `libgit2`
- Created build in `libgit2`
- Built `libgit2`
- Execute `qmake` command

**Branch:**
develop branch

**Solution**
I tried to trace the `libgit2` directory and found no error. It is completely installed.
And then I opened the file `libgit2detect.pri` in `pri` directory inside `fritzing-app` folder, and I found this line:
` exists($$LIBGIT2LIB/libgit2.a) {`

On line _57_ that produced the error. The libgit2 build result has no "libgit2.a", so I create a symbolic link for libgit2.a from libgit2.so using this command
`ln -s libgit2.so libgit2.a`

After creating the symbolic link, I run qmake again, and it succeeds.

KjellMorgenstern:
That's amazing. Dynamic and static library files are usually not interchangeable. 
When building libgit2, run 
cmake -D BUILD_SHARED_LIBS=OFF
and then
cmake --build .
to get a .a file.

haennes:
i got the Same Error too.
Solution:
 move the libgit2.a file into a subfolder called build

Nad42:
Hi all, would like to point out the issue with this code is there is a missing
"LIBGIT_STATIC = false"
on line 18/19 in libgit2detect.pri

soon as you add that program compiles fine if you have all dependencies installed.

PR #3963, opened 2022-07-16
Fix for compiling issues regarding libgit2 etc.

Nad42:
Actual fix for #3595 and #3914; the check for installed packages was not being used hence it always started looking for custom located packages (these locations/files are all seemingly out of date) even though you have them installed properly.


KjellMorgenstern:
$$ marks a system variable in qmake. It is a parameter, not a local variable within the qmake script.

PR #3966, opened 2022-08-01
Fix simulator bugs and a few improvements

failiz:
Adds RGB leds and photocells to the simulator
Shows resistance property after family
Scales and centers smokes comming out of parts
Add reason to not simulate a part

### DIFF
--- a/resources/properties.xml
+++ b/resources/properties.xml
@@ -276,6 +276,7 @@
 
 <property name="power" symbol="W" minValue="0" maxValue="1000000000000" defaultValue="" numeric="yes" editable="yes">
 	<suffix suffix="ResistorModuleID"/>
+	<suffix suffix="PhotocellModuleID"/>
 	<suffix suffix="PotentiometerModuleID"/>
 	<suffix suffix="alps-starter-pot9mm" />
 	<suffix suffix="SparkFun-Electromechanical-TRIMPOT-3MM-2"/>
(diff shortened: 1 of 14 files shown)


KjellMorgenstern:
Last week we have been debugging  a problem. Fritzing sometimes crashes when m_simItems get deleted.
This is already in 0.9.10, and happens not very often:

```
Date:   Sun Dec 6 23:16:30 2020 +0100
    fixed problem with dangling pointers. Now, the simulation items are stored in ItemBase.

```
However, with the new "Not connected / no SPICE data" information display, this gets triggered much more often. We tried QTs deleteLater and other mechanisms, but the only way to avoid the crash seems to leak the memory. I have to look into this again at some point, I think there must be something wrong about how these are handled, not only with the way they are deleted. My next approach would be to make them a fixed element of the item, and only change the  appearance (visibility, smoke, text label... )

failiz:
I do not like this kind of problem and I am not sure what the reason is. I hope you can fix it. 
The good thing is that it is easier to trigger now. Thus, it is easier to check if the solution works...
By the way, I started to work on transitory analysis. It does not seem difficult. I am not sure if I will have to run the simulator twice (one for the DC operating point for multimeters and others and another time for the oscilloscope). I think that ngspice makes an operating point to start the transitory simulation, but not sure if I can get the data easily. In any case, I am quite busy at the moment and it will probably not make it for the next release.

PR #3972, opened 2022-08-21
Do not show hybrid connectors in the tooltip

### LINKED ISSUE #3578: hybrid="yes" view supression breaks hover labels on a bused part
## Current Behaviour

From forum post

https://forum.fritzing.org/t/part-creation-howto-part7-suppress-pcb-layer-howto/7918

This problem I haven’t been able to capture as an image, because it will only happen when you hover over the pins in pcb view (which stops when I exit to take the screen capture) so to see this you will need to load the two parts in to Fritzing and try it. If you hover over the left most pin of J1 (the non hybrid variant 1 part) you will get the message:

NEG-Barrel: negative
barrel connector (4)
2.1mm Barrel jack with
terminal block

as expected.

the right pin gives

POS-Barrel: positive
barrel connector (3)
2.1mm Barrel jack with
terminal block

again as expected. The problem is with the hybrid=“yes” part J2 where the hovering over the left connector gets:

NEG-TERN:negative
connector terminal block
(2)
2.1mm Barrel jack with
terminal block

which is wrong, this is the barrel jack connection pin 4 not the terminal block pin 2. The right pin gives:

NEG-TERN:negative
connector terminal block
(2)
2.1mm Barrel jack with
terminal block

which is also wrong as it is the positive barrel jack pin 3. 
**Build:**
<e.g. Version 0.9.3 (b5c895d32 2016-04-19) Cocoa [Qt 5.12.2]

**Operating System:**
<Windows 7>

**Steps to reproduce:**

download these two parts:

[2.1mm Barrel Jack with Terminal Block-improved.fzpz.zip](https://github.com/fritzing/fritzing-app/files/3965250/2.1mm.Barrel.Jack.with.Terminal.Block-improved.fzpz.zip)

[2.1mm Barrel Jack with Terminal Block-improved-hybrid.fzpz.zip](https://github.com/fritzing/fritzing-app/files/3965252/2.1mm.Barrel.Jack.with.Terminal.Block-improved-hybrid.fzpz.zip)

(remove the trailing .zip to get the actual part files)

- Load both in to Fritzing and drag both in to a sketch

- in pcb view hover over the pins of both parts, as documented above, one part (without hybrid="yes") displays the labels correctly. The hybrid="yes" part displays incorrect labels. 

## Expected Behaviour

both parts should display the correct labels. Given that there is a workaround (not using hybrid to suppress pcb view), the correct fix (as mentioned in the parts format file document) may be to remove the hybrid="yes" method of suppressing views (although the code would need to remain for current parts that use it, so maybe mot a big saving.)

failiz:
This PR fixes #3578. When the top most item is an hybrid connector, we should look for other connectors or items under it to display the info of a non hybrid connector.


failiz:
Wrong branch. Closing this.

PR #4076, opened 2023-10-10
Ac simulation

failiz:
I have been working a few weeks on this and it is time for feedback. I do not want to keep working without knowing that I am going in the right direction and that the code will be merged in the future.

To use this branch, you also need the pull request from the fritzing-parts repository, which contains the oscilloscope, pulse generators and others.

In general, I think it is working quite well. Even long simulations (20s) are quite fast (I have also been working with simple circuits). To use it, you just need to add an oscilloscope and the simulator will simulate a transitory analysis instead a operation point analysis. The transitory analysis simulates until the maximum time that can be shown in any of the oscilloscopes (calculated based on the horizontal position of the oscilloscope and the horizontal scale: time/division). From the time that needs to be shown (in any of the oscilloscopes) to the end time, the simulator generates a fix amount of data (now it is set to 400 SimSteps). This defines the simulation step size. After the simulator is performed, the oscilloscope signals are retrieved and plotted. The status of all the components is shown as at the end of the simulation (LED lights, smokes, etc.).

TODO:

- [ ] Generate noise when a wire is connected to the oscilloscope, but not connected to any other element.
- [ ] Fix issues when the GND is not found. Not finding the ground could cause bad results in the simulator. 
- [x] Create more wavegenerators. I am planning: sine, sawtooth, single pulse, generic pulse, PWM, square. Later, we can add the rest of the buttons of the wavegenerator
- [ ] Wave generators need specific icons. The bb image is too big to be used as an icon. 
- [ ] Now the power supplies will not be sch only parts as there will be bb parts too. Decide where to place them. I propose to add them as section wavegenerators in the sim and the core. 
- [ ] Create a wavegenerator family? Now, I am using the power supply, which allows to swap with dc power supplies and brakes the wiring.
- [x] Add an option in edit->beta features->simulator to allow to set the max points of the simulation or the step time of the simulation. It could be that a circuit cannot be simulated without reducing the timestep.
- [x] Add animation of the transitory circuit (LEDs lighting and signal appearing in oscilloscope) and show sim time.
- [ ] Probably not for now, but add switches driven by a power source. (Basically changes the resistance of the connection depending on the value of the input). This will allow to simulate m
- [ ] Add more examples: 
  - [ ] NAND gates
  - [ ] counters
  - [ ] SR flip-flops
  - [x] 555
  - [ ] full wave rectifier
- [ ] Add the current examples to the list of examples (so, they appear as examples)
- [ ] Homogenize design: The oscilloscope, wavegenerator, dc power supply (made by @vanepp , not sure if he is making a new one more standard) are a bit different between them. It would be great if someone tries to make them to look nicer (same style (color line thickness, etc.), now mine are flat and not in perspective). This is much easier now that there are not published and we cna change them without obsolete them
- [ ] Probably for later: add a control bar to set the time and show the components at that specific time (LED lights, etc.)
- [ ] Probably for later: improve the check of the componets. If a component is damaged at anytime, the component should have a smoke symbol from that moment. 
- [ ] Fix issue when there are several oscilloscopes and one has a big time/div and the other a small one


DanielBarie:
This is great work. Thank you! 
Highly useful for teaching basic circuits/electronics. From a teaching perspective, I'd love to see that integrated in a future release.

failiz:
> This is great work. Thank you! Highly useful for teaching basic circuits/electronics. From a teaching perspective, I'd love to see that integrated in a future release.

Thanks @DanielBarie ! We are trying to merge it into the official code, but it could take a while. Please, feel free to give feedback on the branch. :-)
Great work documenting the building process in https://github.com/DanielBarie/FritzingBuild. I am curious, why is the battery damaged in the second picture? It seems that the voltage is set to 5V and the resistor is 220ohm. The battery should get damaged when there is a short circuit and the current is bigger than 30A (if I remember correctly).

DanielBarie:
Thank you. That's excellent news. 
As to that battery: I was suprised to see that smoke, too. 220 Ohms shouldn't lead to battery overload. Haven't had the time to figure it out, but it was a nice (if surprising) demonstration of the simulator...

DanielBarie:
I just found your paper (Learning Hands-On Electronics from Home:
A Simulator for Fritzing). Even better. I wish I had known about it much earlier.

KjellMorgenstern:
I tried this today. The capacitors in the in the FlipFlop example get a negative voltage and blow up. A real world elco might survive the small voltage. 
image

I flipped them without further thoughts (not 100% sure if that is the intended circuit):
image

Capacitors don't blow up, and the DC analysis ends up (randomly) in one state of the flipflop.

It looks like the transient analysis it not affected by this change, which is a bit surprising. Does it calculate the capacitor charge regardless of polarity ?

src/simulation/simulator.cpp
```diff
@@ -69,7 +69,7 @@ Simulator::Simulator(MainWindow *mainWindow) : QObject(mainWindow) {
 
 	QSettings settings;
 	int enabled = settings.value("simulatorEnabled", 0).toInt();
-	enable(enabled);
+	enable(true);
```
KjellMorgenstern, line 77: I think it is now save to remove that setting entirely.  If there are no objections, I'll to that for 1.0.3

While on it, will it make sense to have a switch between DC and transient analysis?

src/simulation/simulator.cpp
```diff
@@ -194,6 +194,21 @@ void Simulator::simulate() {
...
+			std::cout << "time/div: " << item->getProperty("time/div").toStdString() << " " << time_div << std::endl;
+			double maxSimTime = time_div * 10;
+			QString tranAnalysis = QString(".TRAN %1 %2").arg(maxSimTime/100).arg(maxSimTime);
+			spiceNetlist.replace(".OP", tranAnalysis);
+			//TODO: Handle several oscilloscopes
+			break;
+		}
+	}
+
+
```
KjellMorgenstern, line 249: This probably answers my above question.
However there might be many use cases for transient analysis without an oscilloscope.
We could add a little arrow on the play button, like many other buttons in the menu bar have:

image

src/simulation/simulator.cpp
```diff
@@ -568,6 +589,46 @@ double Simulator::calculateVoltage(ConnectorItem * c0, ConnectorItem * c1) {
...
+		} else {
+			svg.append("L " + QString::number(i, 'f', 3) + " " + QString::number((voltage + v_offset) * vScale + y_0, 'f', 3) + " ");
+		}
+		std::cout << voltage << ' ';
+	}
+	svg += "' fill='red' stroke='black' stroke-width='10'/> \n";
+
+	std::cout << std::endl;
+	return svg;
+
```
KjellMorgenstern, line 756: I think this and other methods might be better placed in items/oscilloscope.cpp or similar.
So, we'd create an Oscilloscope class, that can display probe and comVectors ... in the future it could also hold code to act on buttons and knobs. 

Is there a specific reason it has to be here?

src/simulation/simulator.cpp
failiz: Yes, as soon as there is an oscilloscope, the transitory analysis is used. Before the trans analysis only made sense with the oscilloscope as there was no animation. Since I added the animation, users could use the LEDs in the breadboard to see the effects on their circuits. 
So, yes, maybe it is a good idea to add the trans analysis as an option in the simulate button. However, in that case, how do we specify the simulation time? 
Could you implement the button? The last time I played with adding buttons took me ages to make it correctly.

src/simulation/simulator.cpp
failiz: No, I also thought about that several times. I can create an oscilloscope item and move that code there.

src/simulation/simulator.cpp
failiz: I am OK with removing the enable simulator setting.

failiz:
@KjellMorgenstern , regarding the electrolytic capacitors in the example (flipflop_led_flashing_circuit), I took this circuit from internet. Simulation says that they get -0.76V and that is why they show smoke. If I remember correctly I set a threshold of 0.5V to consider it a failure. I have to try this in reality. Probably, we could just increase the threshold to avoid the smoke. I think that less than 1V is OK.
The spice simulator only has ideal capacitors and they do not have polarity. We just make them electrolytic by showing smoke if the voltage is in reverse and above the threshold. 
In any case, I can see that you wired the BB view in the modification (thanks!) but I do not see that you turned them around. And I do not think they should be turned around as the voltage reaches around 6V and that definitely will destroy the capacitor.

src/simulation/simulator.h
```diff
@@ -64,36 +67,43 @@ public slots:
...
+    void updateCapacitor(unsigned long, ItemBase *);
+    void updateDiode(unsigned long, ItemBase *);
+    void updateLED(unsigned long, ItemBase *);
+    void updateMultimeter(unsigned long, ItemBase *);
+    void updateOscilloscope(unsigned long, ItemBase *);
+    void updateResistor(unsigned long, ItemBase *);
+    void updatePotentiometer(unsigned long, ItemBase *);
+    void updateDcMotor(unsigned long, ItemBase *);
+    void updateIRSensor(unsigned long, ItemBase *);
+    void updateBattery(unsigned long, ItemBase *);
```
KjellMorgenstern, line 89: Please set your editor to use single tabs, not fours spaces, for indentation.

failiz: Done:
image

failiz:
Will you replace all the spaces to tabs?

PR #4095, opened 2023-12-22
Built with Qt-6.6.1 and system boost, spice, quazip

jiapei100:
1. Successfully built with my qt-6.6.1 
2. make boost, spice, quazip from system, rather than local

### DIFF
--- a/phoenix.pro
+++ b/phoenix.pro
@@ -171,7 +171,7 @@ RC_FILE = fritzing.rc
 RESOURCES += phoenixresources.qrc
 
 # Disable this if you have (and want) libgit2 dynamically
-LIBGIT_STATIC = true
+LIBGIT_STATIC = false
 include(pri/libgit2detect.pri)
 include(pri/boostdetect.pri)
 include(pri/spicedetect.pri)
--- a/pri/boostdetect.pri
+++ b/pri/boostdetect.pri
@@ -32,7 +32,7 @@ defined(boost_root, var) {
     message("Using fritzing boost detect script.")
 
     # boost_1_54_0 is buggy
-	BOOSTS = 81
+	BOOSTS = 83
     for(boost, BOOSTS) {
         exists(../../boost_1_$${boost}_0) {
             LATESTBOOST = $$boost
--- a/pri/quazipdetect.pri
+++ b/pri/quazipdetect.pri
@@ -1,59 +1,12 @@
 # Copyright (c) 2021 Fritzing GmbH
+# https://aur.archlinux.org/cgit/aur.git/tree/0001-Quick-Dirty-patch-to-allow-finding-quazip-qt6-on-Arc.patch?h=fritzing
 
-message("Using fritzing quazip detect script.")
+#message("Using fritzing quazip detect script.")
+message("Using custom quazip detect script.")
 
 SOURCES += \
     src/zlibdummy.c \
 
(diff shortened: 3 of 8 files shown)


KjellMorgenstern:
Thanks, but this doesn't fix any open issues, and the build needs to work on *all* systems. That also includes testing and to verify it for regressions.

jklap:
For what it's worth, I've been running with a variation of these changes for a couple of weeks, boost 1.83, Qt 6.6.1, current quazip and haven't seen any issues.

While I see this was closed due to "not fixing any open issues", I suspect it should be considered that staying current on frameworks, such as Qt, should be a key component of maintaining applications -- you stay ahead of security and general bug fixes, often times performance improvements, etc.

To close it with such an abrupt statement, when someone is clearly spending their own time to help maintain and keep the project moving forward is, I think, a disservice to those trying to help. I would hope that some of these changes are being considered internally, heck, could even be using them during normal usage of the app, ie dog-fooding. 

While updating to the latest and greatest Qt is likely not a priority for the primary team maintaining the software (and probably shouldn't be and instead focusing on increasing functionality, bug fixing, etc), someone has done the work; I would see little reason to not consider including it in future releases..... just my thoughts as an engineer who has both supported projects w/PRs and one that has been supported by others PRs.... and if testing end-to-end for regressions is an effort not able to spent by the team then make it available as a "pre-release" for those of us willing to spend time on the bleeding edge for the benefit of others :)

Plus, there are some bug fixes here related to the build process such as the quazip and ngspice include paths that probably should be incorporated as they aren't functionality changes but simply changes to fix the build so it actually works with those libraries as-is vs mucking around with their directory structures.

But there are also changes here that should probably also be rejected-- the change to using OS provided libraries is an example-- maintaining a dedicated build env w/specific versions of boost, ngspice, quazip is paramount for applications like this (ie such that you can define & stick with specific library versions instead of tracking what the OS installs which is often times way behind or sometimes too far ahead). Not to say there can't be a version of the build scripts that used shared/dynamic/system versions of the libraries but to replace the dedicated build env with the system provided version would be a non-starter for me.

KjellMorgenstern:
We are switching from 6.4 to 6.5 with the next release, and had to fix several regression for that. Keep in mind that we also support some older OSes. 

We don't stick with old versions, we upgrade libs almost each release, but sometime we just focus on different things.

Maybe look at the build system from an embedded system view if you want. Not saying it is embedded, but certainly it is an advantage if you don't get bitten by library version xy having a glitch on OS yz, without even knowing that there are people using that combination. And yes, people end up complaining in my inbox about unknown Fritzing builds, or publishing videos with crashes from a botched build, that we cannot possibly reproduce or fix in any way. There are of course ugly things about our .pri files, but they don't affect users in the way many of the verified issues here on github do.

Fiddling with the directory structure is a result from keeping various versions on *all supported OSes* available at the same time during upgrades. We try to minimize that, and review it during upgrades, but then, once it works, it works. 

If you want to help us with Fritzing, and know your way in Qt Creator, SPICE, or something related, please make contact early. We have tried labels like "easy start" or "help wanted", iirc it is directly in the README. I understand they are not most interesting and that this whole repo is not really inviting. Few have ever asked about these labels, so they might be outdated by now.

jklap:
Totally agree on the "embedded system" point of view, nice comparison! I'm always of advocate of controlling you own destiny even while it means additional steps need to be taken.

PR #4110, opened 2024-02-02
Replace deprecated Q_OS_MAC with Q_OS_MACOS

cjmayo:
https://doc.qt.io/qt-5/qtglobal.html#Q_OS_MAC

https://doc.qt.io/qt-6/macos-issues.html#compile-time-flags


cjmayo:
Merged as https://github.com/fritzing/fritzing-app/commit/070cb8eaca99bab20f0a735a847250f0e8b869c5

PR #4181, opened 2024-10-01
Fix arduino platform

failiz:


src/program/platformarduino.cpp
```diff
@@ -83,4 +90,19 @@ void PlatformArduino::upload(QWidget *source, const QString &port, const QString
...
 	process->start(getCommandLocation(), args);
+
+}
+
+bool PlatformArduino::usingArduinoCLI() {
+	//We make an error and analize if we are using the arduino-cli or not
+	auto * process = new QProcess(this);
+	QStringList args;
+	args.append(QString("--board"));
+	args.append(QString(" sssssss"));
```
KjellMorgenstern, line 101: Can we use something like "arduino-cli version" instead of producing a weird error?

failiz: Please, try to give feedback a bit faster. I forgot what I did that and had to spend time checking this again.   

The issue is that I wanted to create an error. You can try to just have an argument "version". If you do that, you must read the standard output and not the error and then it works well if you have arduino-cli. But it will open a new 
 Arduino-IDE and generate a message (not able to open "version" sketch) when the Arduino IDE 1.x is used as programmer. Thus, if you do not create an error and you are using Arduino IDE (1.x), it will launch a new Arduino IDE every time you press program. If there is an error, the Arduino IDE is not launched. 

Not very elegant, but it works well with arduino-cli and Arduino IDE 1.x. Of course, we could drop the support for the Arduino IDE 1.x too, but that is your decision.

KjellMorgenstern: This is fine then, I agree we shouldn't launching the IDE just to probe it.

KjellMorgenstern:
This was merged to 1.0.5, in fd7dda43b and follow up -> Closing

PR #4207, opened 2024-12-21
Update German translation

Atalanttore:


Atalanttore:
@KjellMorgenstern What's wrong with my translation?

KjellMorgenstern:
In general, please contact us before. We use Weblate to manage translations.

Specific to German, we don't need any, since half the contributors speak German. That would of course be different for a huge chunk of text, but the overhead doesn't make sense if it is just a dozen phrases.

In this concrete case, most strings were already outdated when you made the translation (the "Old copper" was a temp solution while fading out that somewhat broken algorithm) .

See also https://fritzing.org/support-us/language-translation/

PR #4249, opened 2025-05-25
Cpp pedantery and Qt 6.9.0 support

JohannesWilde:


KjellMorgenstern:
I did a brief check before the 1.0.4 release, there were several dialogs that broken on Qt 6.7. 
Also, on macos copper fill didn't work anymore. 
We will upgrade from Qt 6.5, but it is not as trivial as bumping the version check in the build file.

JohannesWilde:
Ok, thanks for letting me know.

I also encountered a problem where the ground fill [and copper fill] are not currently working. But this is probably unrelated from the Qt version?

Maybe the other patches are still worthwhile - feel free to cherry-pick, revert or simply close this PR.

PR #4314, opened 2026-06-01
Add Qt 6 build support, panelizer v2, and Gerber preview

landracer:
Qt 6.11.1 build support
panelizer v2
Gerber preview
GCC 14 compatibility fixes


KjellMorgenstern:
Thanks for this PR, but it doesn't really look like you intend to have this merged?

landracer:
> Thanks for this PR, but it doesn't really look like you intend to have this merged?

that is up to you.... the code is there for an extension fritzing should incorporate, but that's just my opinion.  It's something I wanted myself, so regardless my local builds support it now. Up next is an interactive parts editor.  But yea. Totally your call, it's your repo.

KjellMorgenstern:
### Workflow

Without further work also on your side, I fear it won't be possible to even take a small piece out of it. Asking an AI to generate it would be less work than to read the PR. And I would have a trusted workflow, influence on the architecure, and know what was specified.

**Outdated**
The PR is based on a very old version of Fritzing.
This branch 1167 commits behind fritzing/fritzing-app:develop.
There is no way we can we can integrate it like that. Fritzing is already using Qt6 since version 1.0.0, and versions > 6.8 are currently known to break sketches and rendering. I don't think "It compiles" is enough.

Gcc compatibility fixes are probably void by the same reason.

**Size**
The PR is way to big for manual review or testing.
Since AI can churn out somewhat working code in the thousands of lines easily, our limitation isn't really generating/writing the code. The limit we are working on and trying to improve, is, how much code can we integrate savely, tested, with few regression as possible.

### Feature wise
**Panelizer**
Can you describe the use case?
The panelizer can be useful for self etching. But if you use a Fab, better let them optimize the panel, as it will be faster and cheaper. The Fab can optimize space on the panel by combining it with other boards. It is the first time I see the panelizer mentioned in years, there are currently no open issues for it, and the ones closed 4 years ago had been without activity for even more years. So, I guess demand for this feature is low. Still, repairing or replacing Fritzings panelizer with a v2 might be justified, but it needs to be a targeted PR and must be low effort to be feasible.

**Gerber viewer**
Integrating a gerber viewer looks smoother, but it it would be much much simpler to hook in gerbv.
And for praticability: Using gerbv as a reference and second opinion on what Fritzing generates has quite some advantages. No point in reflecting Fritzings code base, just to learn that the Fabs gerber rendering disagrees. gerbv is available cross platform.

I could not find how you specified the Gerber format. Did you feed the AI some standardization document, or did it pull that by itself? 

### Architecture
The PR adds almost 28000 lines, so I just glanced at it briefly, pointing out what catched my eye.

It touches code that goes back to the 1980s and has a "do not touch" comment. Quite bold attempt, don't do it, it is painful.

Panelizer and gerber viewer seem to be quite independent windows, which is nice.

The performance optimization by introducing multithreading doesn't provide measurements, which is a huge no no.
I can imaginge the panelizer triggered performance issues, by inceasing the amount to render practically by factor N=number of boards per panel. But if that assumption is correct, the solution would be run the gerber rendering step first, then just copy this with the panelizer, instead of rendering a huge panel. As soons as you have more boards on the panel than CPU cores, it will be faster, and it would not touch the existing code as much.
This is even more relevant, since we have made significant performance improvements compared to the state this PR is based on. Our use case is running hundreds of Fritzings to hunt bugs, so we didn't chose the brute force multicore path, but algorithmic improvements.

Automated test: It looks like the test added is a placebo. Did I miss anything?


I could go on quite some time, since as you said, this is really a big PR. But I think it would be most interesting to discuss the gerber and panelizer features, and how to incorporate them. Did you consider the gerbv path?

landracer:
Thanks for the detailed and fair review — you were right on the big points, and I've reworked the contribution accordingly. I didn't realize we had so much drift from master to DEV. my appologies. I've respun things a bit. So maybe you can just play around with it in your own dev environment. 

Rebased onto current develop. The original branch was ~1167 commits behind, which made it unreviewable. I've ported the work onto today's develop HEAD. It builds and links cleanly, and the integration footprint into existing code is now tiny and purely additive: a couple of per-layer copperFill/groundFill overloads, one ModuleIDNames constant, the .pri wiring, and the two menu hooks. No changes to core sketch/rendering logic.

They really are independent windows. As you noted, the panelizer and the Gerber preview stand on their own. The panelizer is launched from Production → Panelize… as a self-contained dialog flow (a small start dialog for quantity/blend, then one interactive arrange-and-preview window). The Gerber preview is a non-modal window that opens after a Gerber export so you can sanity-check the output before sending it to a fab. Neither reaches into the editor's internals.

Use case for the panelizer. It's aimed at DIY/self-etching and small-batch workflows: take a finished PCB sketch, tile N copies (optionally blending in other boards) with configurable spacing/separators, arrange them on a panel, and export the combined Gerbers in one pass. The placement backend reuses Fritzing's existing corner-stitching tile engine, so it's not introducing a new algorithm — just driving the layout.

On the Gerber format question — no, I didn't hand the AI a spec and let it guess. The parser and renderer are a clean-room implementation of Ucamco's published Gerber RS-274X specification (https://www.ucamco.com/en/gerber/downloads), plus a small Excellon drill subset. That provenance is documented in the source headers (e.g. gerberparser.cpp, gerberaperture.cpp), and there is explicitly no code taken from gerbv. The parser implements the documented grammar directly: format spec (%FSLAX…), units (%MO), aperture definitions (%AD)/macros, modal interpolation, etc.

On gerbv. I understand the instinct to lean on gerbv rather than maintain a renderer, and I'm not dismissing it. We built this engine on top of several existing projects and have exercised it heavily — it's solid for our purposes, ships no extra runtime dependency, and renders directly into the Qt scene we already use. That said, I think gerbv has real value as an independent second opinion: I'm happy to treat it as a reference oracle for tests (compare our render/parse against gerbv output) rather than a runtime dependency, if that addresses the maintenance concern.

On the "placebo test" critique. Point taken. The branch now includes test_panelizer and test_gerberpreview with fixtures; I'll expand them into real assertions (golden-panel layout and parsed-Gerber geometry) rather than smoke tests, and I'm open to wiring gerbv-based cross-checks as above.

I also dropped the unmeasured multithreading changes and the older sketchwidget edits you flagged, so the diff is now focused on just the two features. Happy to split this into two PRs (panelizer / Gerber preview) if you'd prefer to review them independently.

The branch builds, links, and the unit tests pass on Qt 6.11.1. I've also isolated a second commit — purely forward-compatibility casts (QString::arg()/QChar() on Qt 6.9+) — that's independent of this feature; it's there if you want to use it toward lifting the Qt version cap, but the feature doesn't depend on it and doesn't touch the guard. Everything's on my fork below for you to review directly and pull whatever you find useful.

https://github.com/landracer/fritzing-app/tree/feature/panelizer-gerber-v2-on-develop

lets close this PR and we can use the dialog to push around what I've done up to this point.

KjellMorgenstern:
Yes, please split the commit into PRs feature wise, this is really essential. 
I should note these points should on those specific PRs, but I am already posting them here to reduce iterations:

## Panelizer
Can the panelizer be triggered in CLI ? I'd expect the same output for CLI as well as interactive mode. I saw earlier versions described here used a xml input file: https://forum.fritzing.org/t/v-cut-or-panelizer/5416 . 

What output files does it generate? Specifically, does the panelizer produce a single merged gerber, or some kind of panel defintion file togther with the original gerber?

Are panelizer options stored, do they surive reloading the sketch? Fritzing since around version 1.0.0 has the ability to store project specific settings in the fzz. Just mentioning it, wether storing that makes sense depends on the workflow. I am fine with treating this as an option passed during export, meaning, each time we want a panel, just call the panelizer, no stored setting.

## Gerber Preview

**Beta feature**
It should be a beta feature until it is hardnend. This makes it much easier to integrate.
Anyone using it is warned that it might change without notice.
In preferences, beta feature section, add a checkbox. It is saved to the settings, and enables/disables the preview at runtime.  

**Window managment**
The preview should not get triggered automatically during each gerber export.

Thinking loud:
Adding a distinct menu item in export -> production -> Gerber Preview would allow your flow without breaking others. 
It could also appear in this menu:
image

Another idea would be to extend the gerber export dialog. Currently this is just the systems "save file" dialog.
Adding export options in this dialog would be natural, not surprise users, and be still the same number of steps.
It needs a distinct "save" dialog, we can't use the native dialog anymore, opening cross platform UX considerations, this gets quite complex.

For now, since we treat the gerber preview as a beta feature, lets just add a distinct menu item under File → Export → Production, that hides if the feature is not enabled in preferences. We can revisit the integrated dialog approach later.


**Gerber Preview Testing**
System test:
How would you compare gerbv against the new gerber renderer? I think there should be an option to save a  rasterized export at 1000dpi . 
 That will help testing, and might be neat for anyone who wants to send it to a printeror similar. 
Note 1: I use odiff to compare such images https://github.com/dmtrKovalenko/odiff   (not asking you to integrate odiff, just describing how the rasterized export could be used for testing)
Note 2: Qt is not deterministic when exporting such large images! See the imageToHash() method usage in fritzings gerber export, which we use to mitigate the problem. Observed for large images, but also tiny sketches. Seems to be related to system load, but not CPU specific (reproduced on Intel, AMD, various AWS instances). Larger image size naturally make it more likely to happen, so it will nearly always be triggerd by a large panel at 1000dpi. More insights from a third pair of eyes would be greatly appreciated. See https://github.com/fritzing/fritzing-app/blob/04e5bb0241e8f1de24d0fce9be070041c6d5b68e/src/svg/gerbergenerator.cpp#L384

Unit test:
I didn't look into it this round, would be part of the splitted PR.

## Qt migration
Also split the Qt migration. We're pinned to 6.8.3 due to known rendering regressions in later versions and won't migrate until those are resolved. We'll review the compat fixes independently.

landracer:
> Thanks for the review — this is exactly the kind of feedback that
> makes the feature land cleanly. You're right that this is too big to review
> as one unit, so I'm splitting it into **four independent PRs** that can be
> reviewed and merged in order. Below I answer every question and describe the
> split.

## How I'm splitting it

| # | PR | Scope | Depends on | Risk |
|---|----|-------|-----------|------|
| 1 | **Qt 6.9+ forward-compatibility** | 7 one-line cast fixes only. No version-cap change. | develop | trivial |
| 2 | **Gate connector-error red highlight behind a preference** | Develop-side debug diagnostic that paints the canvas red; now off by default. | develop | trivial |
| 3 | **Gerber Preview (beta feature)** | Standalone Gerber/Excellon viewer, beta-gated, menu item, no auto-open. | develop | medium |
| 4 | **PCB Panelizer wizard** | The panelize wizard + engine; embeds the preview from PR 3. | **PR 3** | large |

PRs 1 and 2 are independent of each other and of 3/4. PR 4 stacks on PR 3
because the panelizer embeds the preview widget.

## Panelizer questions

- **Is there CLI support?** No. The classic Fritzing panelizer was driven by
  an XML input file from the command line; this is a fresh **GUI-only**
  reimplementation reached from *File → Export → for Production → Panelize…*.
  It's a two-step wizard (a tiny start dialog for quantity / rotation / blend,
  then one interactive window with all options + a drag-to-arrange canvas +
  embedded preview). No batch/headless entry point is wired up. The engine
  (`PanelizerEngine`) is already split into a headless `layout()` + `emitPanel()`
  with no UI or file-dialog coupling, so a CLI could be added later without
  touching the UI — but it is explicitly **out of scope** for this PR.

- **What output files does it produce?** A standard fab-ready set written to a
  user-chosen folder: **RS-274X Gerbers + an Excellon drill file** for the
  composited panel (the same `GerberGenerator` path the normal export uses),
  with the panel frame, board-to-board separations (none / V-cut / mouse
  bites), fiducials and tooling holes merged in. It does **not** emit a
  panel-definition sidecar; the output is the gerber set itself.

- **Are panel options stored in the `.fzz`?** No. Panel options live only in
  the wizard session, and the output folder is chosen per-export. Nothing is
  persisted into the project file today. If you'd prefer panel settings to be
  saved with the sketch (via the existing project-properties mechanism), I'm
  happy to do that — but I'd suggest it as a follow-up so PR 4 stays focused.

- **One thing I want to flag for your review:** the panelizer's bin-packing
  uses the corner-stitching `Tile` data structure that used to live in
  `src/autoroute/cmrouter`. That code was removed from `develop`. The branch
  currently re-adds the *whole* `cmrouter` tree, but the panelizer only
  `#include`s `cmrouter/tile.h` and `cmrouter/tileutils.h` — so in PR 4 I'll
  vendor **only** `tile.{cpp,h}` + `tileutils.h` (the geometry primitive) and
  **not** resurrect the deprecated router (`cmrouter.cpp/h`, `DBcell.cpp`,
  `search.cpp`, `search2.cpp`). Let me know if you'd rather I move the
  corner-stitching primitive somewhere neutral instead.

## Gerber Preview — agreed, making it a beta feature

I'll wire it up the way you described, matching the existing Beta Features
pattern in `PrefsDialog`:

- A new **"Enable Gerber preview"** checkbox under *Preferences → Beta Features*,
  saved as `gerberPreviewEnabled` (default **off**).
- The setting is applied **at runtime** via `FApplication::updatePrefs` (same
  live-apply path as the other beta toggles) — no restart needed.
- A distinct menu item **File → Export → for Production → Gerber Preview…**
  (separate from the export action). When the feature is disabled the menu
  item is **hidden**, so users who don't opt in never see it.
- **The preview no longer auto-opens on every Gerber export.** Export just
  exports; the preview is something you open deliberately from the menu. This
  fixes the window-management problem you raised — no surprise windows popping
  up on each export.

## Window management

Addressed by the above: with auto-open removed and the entry point behind an
explicit, hideable menu item, there are no unexpected preview windows. The
preview is non-modal and `WA_DeleteOnClose`, so repeated opens don't leak.

## Testing

- There's a unit test for the gerber parser (`tests/auto/test_gerberpreview`)
  that I extended with an aperture-select regression (the bug where modern bare
  `Dnn*` aperture selects weren't matched, so traces fell back to the 0.1 mm
  default width).
- On your **1000 dpi rasterized export → gerbv → odiff** idea: I agree that's
  the right way to pin down rendering parity, and I see the Qt
  non-determinism concern with large image export
  (`gerbergenerator.cpp:384` `repeatedImageRender` + `imageToHash`). I'd like
  to land the viewer first (behind the beta flag) and add the
  golden-image comparison harness as a follow-up PR so it doesn't block the
  visual feature. The preview is opt-in and clearly labelled beta until that
  harness exists.

## Qt version

I want to be straight with you here, because I think this is the one place we
may see it differently:

- I'm building and running this whole branch on **Qt 6.11.1** and I'm not able
  to reproduce the rendering regressions that the 6.8.3 pin is guarding
  against — panelizer, gerber preview and normal sketch rendering all look
  correct here.
- The forward-compat commit (PR 1) **does not raise the version cap**. It only
  adds explicit casts so existing code keeps compiling cleanly on newer Qt.
  It's safe to merge regardless of what you decide on the pin.
- My honest recommendation is that it's worth re-testing the original
  regressions on current Qt — they may already be fixed upstream, in which
  case the pin is costing you newer-Qt fixes for no benefit. If you can point
  me at the specific regression reports, I'll try to reproduce them on 6.11.1
  and report back.
- To be clear: I'm **not** going to roll the code backward or chase a
  regression I can't reproduce. PR 1 keeps everything Qt-version-neutral, so
  this disagreement doesn't have to block any of the four PRs.

This will take a bit as I had a few days to dice hard into something that has been lurking for 'way to long' SO now that we have a new road-map for all this, it'll all trickle in ASAP.  I'll get to spinning.

KjellMorgenstern:
PR4 Great, looking forward to run tests.

PR3
Please see also my comments in the branch you posted above ( https://github.com/landracer/fritzing-app/tree/feature/panelizer-gerber-v2-on-develop )
I think specifically the Panelizer should use the recent algorithm that is already in use in Fritzing, not resurrected  legacy code. There should be references also in the code, but the algorithm is demonstrated here with excellent documentation: https://github.com/juj/RectangleBinPack

CLI support not needed, but I think we should fade out the old version of the panelizer, as it is probaly not used anymore.

PR2
Connector errors will be refactored in 1.0.8 and partly become a production feature. Of course the annoying red flash (inconsistent net while connectios are manipulated) will be gone.
 
PR1
We tried Qt 6.10 , but had to go back to 6.8.3. We did not take this lightly. What makes this extraordinary expensive is backward compatibility (Qt migration breaks thousands of existing sketches) and cross platform testing. The crash reports from third party Qt 6.10 builds that ended up in our repo showed that people understimate this vastly.

Building against Qt 6.8.3 is 5 minutes + the download time, so why not give it a try? I highly doubt that something essential from a later Qt version is needed for panelizer or gerber preview.

Architecture remark:
OutlierHandler, DRC, Autorouter all use some form of non modal window. 
Can you describe the dependency between panelizer and gerber preview? Is this because of the non modal window (and saving some development time) or is it a genuine requirement from your use case?

Closing in favor of upcomming PR3 and PR4.

landracer:
Thanks for checking this all out — the direction is clear and I've reworked the
branch to match it. Quick summary of where things landed, then answers to your
main questions.

### Panelizer now uses GuillotineBinPack, not the resurrected Tile engine

You were right that the panelizer should use the packer Fritzing already ships
rather than reviving the legacy corner-stitching code. The layout step now runs
on `rbp::GuillotineBinPack` (the same `RectBestAreaFit` / `SplitMinimizeArea`
packer `PCBSketchWidget` uses), working in integer mils. Per-board
`allowRotate90` is honoured: because best-area-fit only rotates a board when it
otherwise won't fit, a forced rotation on a board the user marked non-rotatable
fails the layout instead of silently rotating it.

With the Tile-based layout gone, I removed the whole `cmrouter/tile.{cpp,h}` +
`tileutils.h` tree **and** the dead batch panelizer that only existed to pull it
in (`Panelizer::panelize()`/`inscribe()` had no callers anywhere). The one piece
`FabExporter` actually reuses — `Panelizer::makeSVGs()` — is kept byte-for-byte,
so Gerber output is unchanged. Net effect of that cleanup alone:

```
7 files changed, 127 insertions(+), 3262 deletions(-)
```

So it simplified things by quite a bit, as you predicted — about 3,100 fewer
lines and no more legacy router code in the tree.

### Dependency between the panelizer and the Gerber preview

It's a real, one-directional source dependency, which is why I split them as
PR 3 (preview) → PR 4 (panelizer) rather than shipping them together:

- The preview is **standalone**: `GerberPreviewWidget` / `GerberPreviewDialog`
  under `src/gerberpreview/` know nothing about the panelizer and work on any
  folder of Gerbers (that's PR 3, the opt-in beta viewer).
- The panelizer **embeds** that widget: its preview wizard page
  (`panelizerpages/panelpreviewpage.cpp`) constructs a `GerberPreviewWidget`
  for the live panel preview and pops the standalone `GerberPreviewDialog` for
  "open in full window". The interactive dialog holds one too.

So the panelizer depends on the preview, never the reverse. PR 4 stacks on
PR 3; PRs 1 and 2 are independent of both. If you'd rather review the panelizer
without the embedded preview, I can stub the page behind the same beta flag —
but the natural seam is preview-first.

### Connector-error highlight (PR 2)

You mentioned the connector-error handling is being refactored for 1.0.8 and the
red flash is going away there. If that's landing, PR 2 (which just gates the
existing red-canvas diagnostic behind an off-by-default preference) becomes
redundant — I'm happy to **close it** and let the 1.0.8 refactor supersede it,
or keep it as a one-commit interim toggle until then, whichever you prefer. No
strong opinion from me; it's a stopgap either way.

### Qt version pin

I want to be straightforward rather than push on this. I hear the reasons for
the 6.8.3 pin — backward compatibility (a Qt migration that quietly breaks
thousands of existing sketches is not worth it), the cross-platform retest cost,
and the third-party crash reports on the newer Qt. Those are project-level
risk calls that are yours to make, and none of this feature set asks you to move
the pin.

What I can offer concretely:

- **The features don't need a newer Qt.** To make that easy to verify on your
  blessed toolchain, I put a small helper on its own branch
  (`pr/build-with-qt`, `tools/build-with-qt.sh`) — it builds/runs the tree
  against whatever Qt you point it at (`-q ~/Qt/6.8.3/gcc_64/bin`) by raising
  the `QT_MOST` cap *only for that build* and restoring `phoenix.pro` on exit,
  so it never commits a cap change. `-c` does a configure-only check if you just
  want to confirm qmake accepts it. It's a standalone convenience script, not
  tied to the feature PRs — pull it or ignore it as you like.
- One thing that surfaced while wiring that up: on anything past ~6.5.3,
  **`develop` doesn't compile as-is** — you hit the `QString::arg(Qt::Orientations)`
  / enum-cast errors in `commands.cpp`, `textutils.cpp`, etc. before any of this
  feature code is even reached. PR 1 (`pr/qt6-forward-compat`) is exactly those
  explicit casts, so it's effectively the prerequisite that makes a newer-Qt
  build (and therefore the helper script) produce a binary at all. It **does not
  raise the cap** — it's cap-neutral and safe to merge independent of the pin
  decision; it just lets the existing code keep compiling cleanly forward.
- I'm building/running the whole branch on 6.11 here and can't reproduce the
  rendering regressions the pin guards against. If you can point me at the
  specific regression reports, I'll try to reproduce them on both 6.8.3 and 6.11
  and report back — but I'm not going to roll code backward to chase something I
  can't reproduce. PR 1 keeps everything version-neutral so this doesn't have to
  block any of the four PRs.

Happy to take these in whatever order works for you.

PR #4336, opened 2026-08-29
Drop the random per-session PartFactory folder from parts.db

bmwiedemann:
So that package builds become reproducible.

Without this change, openSUSE's `fritzing-parts-1.0.7` package produced a different result in every build because `/usr/share/fritzing/parts/parts.db` varied.

See https://reproducible-builds.org/ for why this matters.

Reported earlier at https://bugzilla.opensuse.org/show_bug.cgi?id=1239967
_(sorry for leaving ClaudeAI's writeup on #4335. The change is manually reviewed and tested (only build) by me)_

### DIFF
--- a/src/referencemodel/sqlitereferencemodel.cpp
+++ b/src/referencemodel/sqlitereferencemodel.cpp
@@ -1004,6 +1004,8 @@ bool SqliteReferenceModel::insertPart(ModelPart * modelPart, bool fullLoad) {
 				ModelPartShared * mps = modelPart->modelPartShared();
 				if ((mps != nullptr) && (mps->superpart() != nullptr) && mps->superpart()->path().startsWith(prefix)) {
 					bail = false;
+					// drop the random per-session PartFactory folder so parts.db is reproducible
+					path = QFileInfo(path).fileName();
 				}
 			}
 


KjellMorgenstern:
Looks save and makes sense at first glance.
But why was a randomized path chosen in the first place? This is likely an issue with concurrecny. Could a second window now destroy the parts for a first window? Might be outdated, but I don't think that it was just arbitrary.

Parts are not only loaded from the prebuild DB, but could also be 3rd party files imported by the user. There is also the edge case of running different versions in parallel. We re-organized storing parts in 1.0.7, to reduce the most common cases of filename clashes, but this introduced bugs when running 1.0.6 and 1.0.7 at the same time (fixed in 1.0.8).

So a possible scenario : First windows is closed, cleans up folders. Second window keeps on working fine. When the user chooses to save his work, files get collected, but now can't be found anymore. Some other places of the code regenerate the subpart on the fly, but it was to late, and a corrupted sketch gets stored, error not noticed, overwriting the previously saved copy. Not saying that it would work that way, just drawing a pesimistic scenario that might be worth checking.

bmwiedemann:
The random path is needed at some point to avoid one process clobbering files of another. But if I understood it correctly, this is short-lived and gets cleaned up, so keeping this random path around for later has no use, because the files will no more be there.

